PR #528, opened 2019-03-28
New hyperplane implementation

ts4051:
Will implement the new hyperplane system in this branch.

Creating a PR now to ease visibility for others during the process. Not yet requesting a merge.


pisa/utils/hyperplane.py
```diff
@@ -0,0 +1,628 @@
...
+
+
+
+class SysParam(object) :
+    '''
+    A class defining the systematic parameter in the hyperplane
+    Use constructs this by passing the functional form (as a function)
+    '''
+
+    def __init__(self,name,nominal_value,func_name,initial_fit_params=None) :
```
jllanfranchi: can you make `initial_fit_params` a dict instead of a list so the user explicitly sets param values by name, or does that make things tricky / worse elsewhere? I always prefer named things, but if it makes other code difficult/impossible/ugly, then I understand making these a simple list of values.

pisa/utils/hyperplane.py
```diff
@@ -0,0 +1,628 @@
...
+
+        # Store basic members
+        self._name = name
```
jllanfranchi: make all these public unless you have good reason for them to be private

pisa/utils/hyperplane.py
```diff
@@ -0,0 +1,628 @@
...
+
+    def get_func(self,func_name) :
+        '''
+        Find the function defining the hyperplane functional form.
+
+        User specifies this by it's string name, which must correspond to one 
+        of the pre-defined functions.
+        '''
+
+        assert isinstance(func_name,basestring), "'func_name' must be a string"
```
jllanfranchi: Something to get in the habit of doing is using `from six import string_types` and use `isinstance(func_name, string_types)` instead of `basestring` with the idea in mind that we're gonna have to be python3 compatible within a year or so. We need to make this change throughout all of PISA, but start thinking for new code in terms of making it py3 compatible.

pisa/utils/hyperplane.py
```diff
@@ -0,0 +1,628 @@
+import numpy as np
```
jllanfranchi: always always use
```from __future__ import absolute_import, division, print_function```
at the top of your imports; see comment below about py3 compatibility

PR #531, opened 2019-05-16
Data generator fix

philippeller:


philippeller:
I tested this in grid and events mode to make oscillogramms and both worked

PR #532, opened 2019-05-16
Adding stages to run on data release CSV files

philippeller:


pisa/stages/discr_sys/pi_hyperplanes.py
```diff
@@ -131,41 +131,57 @@ def __init__(
     ):
         # -- Read fit_results_file and extract necessary info -- #
 
-        fit_results = from_file(fit_results_file)
+        if fit_results_file.endswith('.json'):
```
jllanfranchi: We handle more than just files with ".json" suffixes (e.g. json.bz2, to start) . If csv is how you identify release files, check for csv, handle it, and your "else" condition should be to use `from_file`

philippeller:
- I append those with "hist" because this data in the CSV is in the form of hists
- the stages are sufficiently different to make them into 3 separate ones
- the CSV reader from pandas works just fine, the import only happens when you use those stages, so it doesn't need to be a hard requirement
- hyperplanes still works for me, if someone else can test it too, please do
- I will make the small change of checking the filename ending

jllanfranchi:
If you think Pandas is a good requirement to have, even if it's optional, you need to include that fact in the documentation.

philippeller:
ok, made all changes....gonna merge a little later if that's ok

PR #533, opened 2019-06-13
Add PISA PiStage for earth absorption, minor additions to Map

FactSphere:
I created a new PiStage that calculates earth absorption for neutrinos. The stage first calculates the depth of a water column that would have an equivalent number of scattering targets using the layers utility that is also used for oscillation matter effects. The cross-sections are taken from a ROOT file the path to which has to be given as an argument. This requires the installation of PISA with ROOT support. 
The stage supports binned calculation mode, which is highly recommended for speed. The xsection calculation is not optimized for parallel execution, but it's reasonably fast in binned mode and it only needs to be done once during setup, so the speed doesn't matter that much.
There are also two minor additions to the Map class: Putting labels in bins showing the bin content and dividing one map by a reference map to show, for example, oscillated/non-oscillated histograms.


pisa/core/map.py
```diff
@@ -733,6 +741,14 @@ def plot(self, symm=False, logz=False, vmin=None, vmax=None, backend=None,
 
         X, Y = np.meshgrid(x, y)
         pcmesh = ax.pcolormesh(X, Y, hist.T, **pcolormesh_kw)
+        if not binlabel_format is None:
+            X_mid = np.true_divide(X[1:, 1:] + X[1:, :-1], 2)
```
philippeller: Binning objects provide `midpoints`, that are nicely centered for both linear and log-spaced binnings, so you may want to switch to using those (which will also save a few lines of code here)

pisa/stages/abs/pi_earth_absorption.py
```diff
@@ -0,0 +1,244 @@
...
+The stage calculates first the depth of a water column that is mass-equivalent
+to the path traversed by the neutrino through the earth. This is done
+using the same Layers module that is also used for oscillation.
+The survival probability is then calculated from the average cross-section
+with protons and neutrons.
+"""
+from __future__ import absolute_import, print_function, division
+
+import numpy as np
+import ROOT
```
philippeller: Since this needs ROOT, we should make sure it is declared as an optional dependency in the INSTALL.md / setup.py

pisa/stages/abs/pi_earth_absorption.py
```diff
@@ -0,0 +1,244 @@
...
+
+    """
+    Na = 6.022E23 # nuclei per cm^(-3)
+    # The molar mass is 1 g for pure nuclei, so there is a hidden division
+    # here by 1 g/mol. Also, int_rho is the depth of a mass equivalent
+    # water column, where water has the density of 1 g/cm^3.
+    # So the units work out to:
+    # int_rho [cm] * 1 [g/cm^3] * xsection [cm^2] * 1 [mol/g] * Na [1/mol] = [ 1 ] (all units cancel)
+    out[0] = np.exp(-int_rho[0]*xsection[0]*Na)
+
```
philippeller: add newline character

pisa/stages/abs/pi_earth_absorption.py
```diff
@@ -0,0 +1,244 @@
...
+    ----------
+    int_rho : scalar
+        depth of mass equivalent water column in cm
+    xsection : scalar
+        cross-section per nucleon in cm^2
+    out : scalar
+        Result is stored here
+
+    """
+    Na = 6.022E23 # nuclei per cm^(-3)
```
philippeller: I believe constants like this are avilable in pint, need to check (but that's a detail)

pisa/stages/abs/pi_earth_absorption.py
```diff
@@ -0,0 +1,244 @@
...
+        # what are keys added or altered in the calculation used during apply
+        # The weights are simply scaled by the earth survival probability
+        output_calc_keys = ('survival_prob', )
+        # what keys are added or altered for the outputs during apply
+        output_apply_keys = ('weights', )
+
+        # init base class
+        super(pi_earth_absorption, self).__init__(data=data,
+                                                  params=params,
+                                                  expected_params=expected_params,
```
philippeller: seems like there is some formatting gone wrong here (?)

pisa/core/map.py
FactSphere: In this case, x and y have already been linearized in lines 723 to 727, so they are always centered the way this is written right now already without any extra work.

pisa/stages/abs/pi_earth_absorption.py
FactSphere: I found the place in INSTALL.md where ROOT is mentioned as an optional dependency, but I didn't find anything in setup.py. What do you mean that it should also be listed there?

philippeller: that's probably fine if it is not in setup.py, as ROOT might be not pip installable. In INSTALL.md it currently says:
"ROOT >= 6.12.04 with PyROOT Necessary for xsec.genie and unfold.roounfold services, and to read ROOT cross section files in the crossSections utils module. Due to a bug in ROOT's python support (documented here https://github.com/IceCubeOpenSource/pisa/issues/430), you need at least version 6.12.04 of ROOT." So there it should be mentionned that is also needed for your stage, and if you need a specific version etc

pisa/core/map.py
philippeller: actually, `weighted_centers` is what I meant (https://github.com/IceCubeOpenSource/pisa/blob/master/pisa/core/binning.py#L821). this would center your binlabel nicely in the middle of the bin also when bins are logarithmic. The average you're doing right now places it not visually in the middle was my guess

pisa/core/map.py
```diff
@@ -733,6 +741,14 @@ def plot(self, symm=False, logz=False, vmin=None, vmax=None, backend=None,
 
         X, Y = np.meshgrid(x, y)
         pcmesh = ax.pcolormesh(X, Y, hist.T, **pcolormesh_kw)
+        if not binlabel_format is None:
```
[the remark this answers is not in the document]
FactSphere: fixed.

pisa/stages/abs/pi_earth_absorption.py
```diff
@@ -0,0 +1,244 @@
...
+        for container in self.data:
+            vectorizer.multiply(container['survival_prob'], out=container['weights'])
+            
+    def calculate_xsections(self, flav, nubar, energy):
+        '''Calculates the cross-sections on isoscalar targets.
+        The result is returned in cm^2. The xsection on one 
+        target is calculated by taking the xsection for O16
+        and dividing it by 16. 
+        '''
+        if flav == 0: pdg = 12
```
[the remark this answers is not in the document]
FactSphere: Thank you for the suggestion, that is indeed a much more elegant way to do it and I implemented it in my commits today.

pisa/stages/abs/pi_earth_absorption.py
```diff
@@ -0,0 +1,244 @@
...
+from pisa.utils.numba_tools import WHERE
+from pisa.utils import vectorizer
+from pisa.utils.resources import find_resource
+from pisa.core.map import Map
+
+class pi_earth_absorption(PiStage):
+    """
+    earth absorption PISA Pi class
+
+    Paramaters
```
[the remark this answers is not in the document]
FactSphere: Okay, thank you for clearing that up. Many other stages did it that way, that's why I did it, too. I have now made the parameters arguments to the __init__ function, as you suggested.

pisa/core/map.py
```diff
@@ -518,6 +519,8 @@ def compare(self, ref):
         assert isinstance(ref, Map)
         assert ref.binning == self.binning
         diff = self - ref
+        with np.errstate(divide='ignore', invalid='ignore'):
```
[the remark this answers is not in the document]
FactSphere: Done in latest commit.

pisa/stages/abs/pi_earth_absorption.py
FactSphere: I'm not sure what you mean, but I changed the formatting now. It it better?

pisa/stages/abs/pi_earth_absorption.py
FactSphere: Hm I didn't find it...

pisa/stages/abs/pi_earth_absorption.py
FactSphere: I hope that is correct now?

pisa/stages/abs/pi_earth_absorption.py
FactSphere: Okay, I moved the import to the setup function and mention the stage in INSTALL.md in the line you quote.

pisa/core/map.py
FactSphere: They are placed in the middle on linear and logarithmic stages as it is now, because x and y are always made linear before it gets to this point.
`        if to_plot.binning.dims[0].is_log:
            xticks = 2**(np.arange(np.ceil(np.log2(min(x))),
                                   np.floor(np.log2(max(x)))+1))
            x = np.log10(x)
        if to_plot.binning.dims[1].is_log:
            yticks = 2**(np.arange(np.ceil(np.log2(min(y))),
                                   np.floor(np.log2(max(y)))+1))
            y = np.log10(y)`

philippeller:
ok, so looks like there are only 2 small changes to be done and then this is good to be merged

FactSphere:
Oh goodness... I just realized that apparently while testing, my edits were overshadowed by an old version of the file that I had somewhere in the fridge? And now, testing again with the correct version as it is now, the apply() function doesn't work anymore and I'm at a complete loss as to why that is. Do you guys have an idea how it's possible that the key "survival_prob" is not in the container? Because I do set it in the setup function and unlink the containers again, but for some reason it's not there? Do I need to put .get(WHERE) somewhere? I must admit that I really don't fully understand the structure behind those containers yet...

`---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-2-854f5804ad71> in <module>()
     17 # that is the sum of all categories and pipelines. This is what we want to compare
     18 # to data.
---> 19 templates = template_maker.get_outputs(return_sum=False)
     20 
     21 # Note that when we use the return_sum=False the returned templates are a list

/afs/ifh.de/group/amanda/scratch/trettin/pisa/pisa/core/distribution_maker.pyc in get_outputs(self, return_sum, sum_map_name, sum_map_tex_name, **kwargs)
    123 
    124         """
--> 125         outputs = [pipeline.get_outputs(**kwargs) for pipeline in self] # pylint: disable=redefined-outer-name
    126         if return_sum:
    127             if len(outputs) > 1:

/afs/ifh.de/group/amanda/scratch/trettin/pisa/pisa/utils/profiler.pyc in profiled_func(*args, **kwargs)
    123         try:
    124             start_t = time()
--> 125             return func(*args, **kwargs)
    126         finally:
    127             end_t = time()

/afs/ifh.de/group/amanda/scratch/trettin/pisa/pisa/core/pipeline.pyc in get_outputs(self, inputs, idx, return_intermediate)
    343             try:
    344                 logging.trace(">>> BEGIN: {}.run(...)".format(name))
--> 345                 outputs = stage.run(inputs=inputs) # pylint: disable=redefined-outer-name
    346                 if return_intermediate:
    347                     if outputs is None:  # e.g. for PISA pi

/afs/ifh.de/group/amanda/scratch/trettin/pisa/pisa/core/pi_stage.pyc in run(self, inputs)
    262             raise ValueError("PISA pi requires there not be any inputs.")
    263         self.compute()
--> 264         self.apply()
    265         return None
    266 

/afs/ifh.de/group/amanda/scratch/trettin/pisa/pisa/utils/profiler.pyc in profiled_func(*args, **kwargs)
    123         try:
    124             start_t = time()
--> 125             return func(*args, **kwargs)
    126         finally:
    127             end_t = time()

/afs/ifh.de/group/amanda/scratch/trettin/pisa/pisa/core/pi_stage.pyc in apply(self)
    247         # else:
    248         self.data.data_specs = self.output_specs
--> 249         self.apply_function()
    250 
    251         if self.mode == "BBE":

/afs/ifh.de/group/amanda/scratch/trettin/pisa/pisa/utils/profiler.pyc in profiled_func(*args, **kwargs)
    123         try:
    124             start_t = time()
--> 125             return func(*args, **kwargs)
    126         finally:
    127             end_t = time()

/afs/ifh.de/group/amanda/scratch/trettin/pisa/pisa/stages/absorption/pi_earth_absorption.py in apply_function(self)
    197     def apply_function(self):
    198         for container in self.data:
--> 199             vectorizer.multiply(container['survival_prob'].get(WHERE), out=container['weights'])
    200 
    201     def calculate_xsections(self, flav, nubar, energy):

/afs/ifh.de/group/amanda/scratch/trettin/pisa/pisa/core/container.pyc in __getitem__(self, key)
    358                 return self.get_scalar_data(key)
    359             except KeyError:
--> 360                 raise KeyError('"%s" not found in container "%s"'%(key,self.name))
    361 
    362     def __setitem__(self, key, value):

KeyError: '"survival_prob" not found in container "nue_cc"'`

philippeller:
what are your input, calc and output settings?

FactSphere:
> what are your input, calc and output settings?

input_specs = events
calc_specs = calc_grid
output_specs = events

FactSphere:
`calc_grid.order = true_energy, true_coszen`
`calc_grid.true_energy = {'num_bins':200, 'is_log':True, 'domain':[1., 1000] * units.GeV, 'tex':r'E_{\rm true}'}`
`calc_grid.true_coszen = {'num_bins':100, 'is_lin':True, 'domain':[-1,1], 'tex':r'\cos\,\theta_{z,{\rm true}}'}`

philippeller:
ok, and you're running on CPU?

FactSphere:
<< PISA_FTYPE env var is defined as: "double"; PISA_TARGET env var is defined as: "parallel"; PISA is running in double precision (FP64) mode; numba is running on CPU (multicore) >>

FactSphere:
I get the same error running on single core. It's getting late for me, I'll come back to it tomorrow.

philippeller:
ok....i'm looking through your code and it all look correct....so not yet sure what's happening. You correctly declared `survival_prob` as an output_calc_key

FactSphere:
I am very puzzled by this atm, maybe when I look at it tomorrow it turns out to be something stupid about how I run the stage. I hope.

philippeller:
fair enough, let me know if there are still problems tomorrow.....but i'm confident it might be something simple, as you said it was working correctly earlier

FactSphere:
Okay, I figured out what the problem is, but I'll need your help to fix it: 
Since all of my parameters are only used during setup of the stage, the `params` argument is left empty and `expected_params = ()`. But that is what causes the above error to appear, because the calculation keys are somehow not added to the containers if there are no params given. Is that intended PISA behavior for a stage? Because essentially what I'm trying to do here, to have a stage that does not have variable parameters and only runs once during setup, is not possible. I'd have to declare at least one parameter in `expected_params` and give it in the 'params' argument.

jllanfranchi:
One possibility is that it expects an iterable; `()` are parentheses interpreted as delineating "no operation" while an empty tuple is to be written as
```python
expected_params = (,)
```
or maybe you would prefer to write out
```python
expected_params = tuple()
```

Worth trying out quickly to see if this fixes things...

(Side note: You can use triple back-ticks and optionally specify programming language to add syntax highlighting in Github comments for quoting code, error messages, etc.; I updated some of your comments above already. See [here](https://help.github.com/en/articles/creating-and-highlighting-code-blocks) for more on that, and [here](https://help.github.com/en/articles/basic-writing-and-formatting-syntax) for general syntax guide.)

FactSphere:
The empty tuple doesn't fix the issue, neither does setting params to "{}" rather than None.

FactSphere:
Okay, I found a fix for the PiStage base class: The compute function has to run even if there are no parameters given but there _are_ `output_calc_keys` given. Otherwise, the calculation containers are never created and the stage fails to function. IMO that is a more forgiving behavior.

PR #534, opened 2019-07-10
added pi stage for genie xsec sys

wym109:
Added 4 BY parameters, DIS diff xsec error, high W hadronization error, DIS A-scaling error


pisa/stages/xsec/pi_genie_sys.py
```diff
@@ -0,0 +1,230 @@
...
+                genie_cv2uby,
+                container['GENIE_linear_fit_CV2uBY'].get(WHERE),
+                container['GENIE_quad_fit_CV2uBY'].get(WHERE),
+
+                out=container['weights'].get(WHERE),
+            )
+
+            # Differential xsec systematic
+            diff_dis = nu_diff_dis
+            diff_norm = 1.- (1.6525*diff_dis)
```
jllanfranchi: There should be no non-obvious "magic" numbers in code without comment. Please provide some explanation what this is, where it comes from (a specific reference), just something so I can check if `1.6525` is correct and understand where it comes from.

Maybe it would suffice to put a reference at the top of the file in the module-level docstring, but just make sure it's something that makes it at least somewhat easy for others to verify all the numbers in this code.

pisa/stages/xsec/pi_genie_sys.py
```diff
@@ -0,0 +1,230 @@
...
+
+            if 'bar' in container.name:
+                diff_dis = nubar_diff_dis
+                diff_norm = 1.- (1.8073*diff_dis)
```
jllanfranchi: Probably the same source for this magic number as the above...

pisa/stages/xsec/pi_genie_sys.py
```diff
@@ -0,0 +1,230 @@
...
+            output_apply_keys=output_apply_keys,
+        )
+
+        assert self.input_mode is not None
+        assert self.calc_mode is None
+        assert self.output_mode is not None
+
+    @profile
+    def apply_function(self):
+        genie_ma_qe = self.params.Genie_Ma_QE.m_as('dimensionless')
```
jllanfranchi: Avoid so much copy-pasting by defining somewhere the arguments to your function and then you can use a dictionary comprehension. E.g., since you've already defined the things you need elsewhere, you don't have to have so much effectively redundant code:

```python
kwargs = {name.lower(): getattr(self.params, name).m_as("dimensionless") for name in self.expected_params}
for container in self.data:
    for name in self.input_apply_keys:
        kwargs[name] = container[name].get(WHERE)
    apply_genie_sys(out=container['weights'].get(WHERE), **kwargs)
```

philippeller:
There hasn't been any dev on this PR half a year. @wym109 is this still happening? I'll close it otherwise

wym109:
Not really, I'll come back to this early next year.

philippeller:
Ok, thanks @wym109, good to know! I'll leave this open then

atrettin:
After a conversation with @wym109 we came to the conclusion that his PR is no longer necessary and can be closed. The added systematics have a negligible effect.

philippeller:
Ah ok, thanks for letting us know!
Feel free to re-open anytime if things change

PR #535, opened 2019-07-10
Added stage to shift and/or scale pid (checked out from personal oscnext_santa branch)

LeanderFischer:
Added a stage that can be used to shift pid values by a bias value or scale them with a factor. This can be used to modify the pid cut value used in an analysis without having to change the binning config file.


jllanfranchi:
Looks good to me! I'll have @philippeller take a look and merge when he's able to.

PR #540, opened 2019-09-09
Corrected references for equations. Changed likelihood names.

austinschneider:


philippeller:
@a-tahmid are these changed to your code ok? If yes let me know and i can merge these trivial changes directly

austinschneider:
The changes should not alter the functionality in any way. It's a correction to the references as neither of these likelihoods are actually described in arxiv:1712.01293. What I have labelled now as mcllh_mean is described in arxiv:1901.04645 and provides a good numerical approximation to eq. 26 of arxiv:1712.01293. Eq. 20 of arxiv:1712.01293 is another likelihood entirely and does not provide a good numerical approximation to eq. 26 of arxiv:1712.01293. Additionally, since both of the likelihoods were first described in arxiv:1901.04645 I have changed the labeling to more closely reflect the notation used in the paper. This is sort of a subtle issue as the derivation of eq. 20 in arxiv:1712.01293 is a little hard to follow with the chosen notation of that paper, but in fact the discovery of LMean as described in arxiv:1901.04645 was the motivation for writing arxiv:1901.04645.

austinschneider:
I can provide some code for eq. 20 in arxiv:1712.01293 if there is interest, but it seems to perform worse than other methods.

philippeller:
@austinschneider Yes I saw the changes are only about naming / reference.....just as a courtesy I'd like to give @a-tahmid a chance to acknowledge these as he is the one who wrote the PISA implementation.
If we won't hear back soon I'll go ahead and merge it as is

a-tahmid:
I've taken a look at the changes, and they make sense. I had some suggestions:

Since we're changing the names to reflect the nature of the equation, instead of `mcllh_mean` and `mcllh_eff` we should perhaps name it `pgllh_mean` and `pgllh_eff` where 'pg' stands for 'poisson-gamma'. Also, under the reference for LMean, I think it'd be a good idea to add that LMean is a good analytic approximation to eq 26 from arxiv:1712.01293.

PR #545, opened 2019-11-08
Minor additions/fixes to master

ts4051:
- Fix to atmospheric muon stage (consistent use of variable to specify input name)
- Added NuFit v4.0 global fit oscillation parameter values in a config file


philippeller:
Looks good to me, and I had a quick glance at the nufit4 values, they seem correct

PR #546, opened 2019-11-08
Some minor additions/fixes to master

ts4051:
- Added NuFit v4.0 osc params (cfg file)
- Fix to atmospheric muon stage, not uses a consistent method to get the correct arrays from the input data


philippeller:
Looks good, I quickly checked nufit4 values and they look correct

PR #554, opened 2019-12-17
Hypersurface developments

atrettin:
This fork contains my developments extending the hypersurface code. It introduces all the new options I talked about in the low energy call. The most substantial technical change is the move from curve_fit to scipy.minimize with an explicitly defined chi-square loss function. This introduces the ability to set priors on the coefficients. The covariance matrix returned by curve_fit is replaced by the inverse hessian returned by minimize.
Other changes are the option to set log-mode, bounds on parameters, options to fix the intercept to its initial value, and an option to ignore empty bins.


philippeller:
Changes look good to me, but did not do any testing, just superficial reading of code.

PR #555, opened 2019-12-19
MCEq Barr PISA stage

ts4051:
A PISA stage for generating both nominal and systematic variations to the atmospheric neutrino flux (both handled in a single stage), using splines generated from MCEq. 

To use this in a pipeline, replace both the previous `flux.pi_honda_ip` and `flux.pi_barr_simple` stages with `flux.pi_mceq_barr`. An example cfg file is here: https://github.com/IceCubeOpenSource/fridge/blob/master/analysis/oscNext_std_osc/settings/pipeline/stages/neutrino_flux_stage.cfg

Spline files are found here (generated for combinations of primary CR flux models and hadronic models, all assuming a proton primary): https://github.com/IceCubeOpenSource/fridge/tree/master/analysis/common/data/flux

Cosmic ray spectrum uncertainties are modelled via varying the power law. Hadronic uncertainties are modelled by varying the Barr parameters. Kaon parameters have dedicted K and Kbar parameters. Pion parameters have just pi (not pibar) parameters + an overall pi/pibar ratio parameter (this is also how the Barr 2006 paper did their study).

Note that the old `pi_simple` stage has been renamed `pi_barr_simple` for clarity.


ts4051:
Note that MCEq has undergone major changes recently, and the script to produce the splines is producing different results to before (even after having been updated to support the new version). This will be investigated, but does not block the merge of the PISA stage since we already have all the splines we need.

ts4051:
Some modifications to flux variable names in the old flux stages were made to be more common between the old and new (easier to make comparison), and to easier support modularity.

pisa/utils/fileio.py
```diff
@@ -401,7 +401,15 @@ def from_cfg(fname):
 def from_pickle(fname):
     """Load from a Python pickle file"""
     try:
-        return pickle.load(file(fname, 'rb'))
+        # Open the file (binary)
+        f = open(fname, 'rb')
+        try :
+            # Try standard pickle load
+            return pickle.load(f)
+        except :
```
philippeller: can you add the specific exception? Such that only the one with the wrong encoding is handled
If there is another exception, like missing file, it should be thrown

pisa/utils/hyperplane.py
```diff
@@ -0,0 +1,1416 @@
+import os, sys, collections, copy, inspect
```
philippeller: This file should not be in this PR

pisa/stages/data/grid.py
```diff
@@ -0,0 +1,107 @@
+"""
```
philippeller: Is this any different from using the `pisa/stages/data/toy_event_generator.py` with mode `binned` and `random=False`?
Should check if we can remove code duplication here

pisa/stages/flux/pi_mceq_barr.py
```diff
@@ -105,267 +186,339 @@ def __init__(self,
...
+        barr_z_K = self.params.barr_z_K.value.m_as('dimensionless')
+        barr_w_antiK = self.params.barr_w_antiK.value.m_as('dimensionless')
+        barr_x_antiK = self.params.barr_x_antiK.value.m_as('dimensionless')
+        barr_y_antiK = self.params.barr_y_antiK.value.m_as('dimensionless')
+        barr_z_antiK = self.params.barr_z_antiK.value.m_as('dimensionless')
+
+        # Map the user parameters into the Barr +/- params
+        # pi- production rates is restricted by the pi-ratio, just as in arXiv:0611266
+        gradient_params_mapping = collections.OrderedDict()
+        gradient_params_mapping["a+"] = barr_a_Pi
```
philippeller: would be less error prone code is we did first directly `gradient_params_mapping["a+"]` = self.params.barr_a_Pi.value.m_as('dimensionless')

and the followed in a loop like `for name in self.barr_param_names: gradient_params_mapping[name + '-'] = self.antipion_production(gradient_params_mapping[name + '+'])` or something along these lines

pisa/stages/flux/pi_mceq_barr.py
```diff
@@ -105,267 +186,339 @@ def __init__(self,
...
-               barr_h_pos, barr_h_neg, barr_h,
-               barr_i_pos, barr_i_neg, barr_i,
-               barr_w_pos, barr_w_neg, barr_w,
-               barr_y_pos, barr_y_neg, barr_y,
-               barr_z_pos, barr_z_neg, barr_z,
-               ):
-    '''
-    calculate the modification factor for the flux
+            # Check for negative results from spline
+            negative_mask = container['nu_flux'].get(WHERE) < 0
```
philippeller: these are CPU operations as defined here. They should come after `container['nu_flux'].mark_changed(WHERE)` and then be executed on the host like: `negative_mask = container['nu_flux'].get('host') < 0` etc, followed by `.mark_changed('host')`

pisa/utils/hyperplane.py
ts4051: oops, removed

pisa/utils/fileio.py
ts4051: done

jllanfranchi:
We're abandoning python 2 entirely for pisa, we don't want to have to support both (lots more code to write, lots more tests to run, lots more maintenance, etc) so change fileio to just load pickles the python 3 way (unless this means old pickles created in py2 can't be loaded from py3).

pisa/stages/flux/pi_mceq_barr.py
ts4051: Done

pisa/stages/flux/pi_mceq_barr.py
ts4051: done

philippeller:
@jllanfranchi yes, this is about reading pickle files _created_ in py2 from within py3

pisa/stages/data/grid.py
ts4051: Using `toy_event_generator.py` in binned mode just produces a binned version of the generated data (which follows a power law spectrum) I believe. `grid.py` produces an actual grid, e.g. one event per grid point. There are also complications with bin edges vs bin centers I believe.

ts4051: So I think these two provide different behaviour right now. `grid.py` is good for plotting flux, oscillograms, etc.

jllanfranchi:
Okay, very well then. Carry on.

pisa/stages/data/grid.py
philippeller: ah yeah, I think you're right

ts4051: I have a feeling that maybe `toy_event_generator.py` did produce grid-like outputs in the past but no longer does, but maybe I'm imagining that. In any case I think these are two distinct and both valid cases.

ts4051: Did notice that the docs were out of date though in this stage, have updated now.

PR #556, opened 2020-01-21
Fix memory leak in pi_globes stage

atrettin:
I discovered and fixed a memory leak in the stage wrapping the "globes" sterile oscillation calculator.


jllanfranchi:
Looks good to me! Note this should be merged via "squash and merge"... @philippeller can you provide a quick double check of the code? (I'm adding you as a reviewer.)

jllanfranchi:
Hi Alexander, it looks like you are continuing to develop on your `master` branch and pushing changes that are appearing here. This isn't necessarily a problem, but I do not know if you had intended for your changes to show up in this PR.

So please let us know if you want us to merge right now, wait for you to make more changes, or merge at a previous commit.

In the future:

1. Always create a new branch locally to develop _anything_ (a new feature, make a bugfix, etc.); I advise against developing on your own `master` branch: `git checkout master ; git checkout -b new_feature_xyz`
2. Once you are happy with your task and have committed your changes (`git commit <...>` however many times you do this is fine), make sure to push your code to github (`git push --set-upstream origin new_feature_xyz`), and then submit a PR from that new branch to `icecubeopensource/master` using github's web interface to create the PR
3. Once we merge the feature / bugfix into `icecubeopensource/master`, you can delete the branch you created to develop that feature: `git branch -d new_feature_xyz` (that deletes it locally; you can delete it on github, too, via the web interface)
4. You then can merge from `icecubeopensource/master` into your own `master` branch and the two should be ~ equivalent: `git checkout master ; git merge icecubeopensource/master` (assuming you have `icecubeopensource` defined as a remote wherever you work with your git; you can make the name shorter if you like but still point to that repo)

Steps 1 and 4 are useful since sometimes it makes sense from our end to "squash and merge" your changes, which merges all of your commits into a single commit to `icecubeopensource/master`, we have one single commit which adds whatever feature / fixes whatever bug you sought to add / fix. Then within `icecubeopensource/master` it's easy to checkout the "before the feature/fix" and "after the feature/fix" in case we need to debug code in that way.

If, however, you developed on your `master` branch and made all the commits there while we squashed then merged, then it creates issues with you merging from `icecubeopensource/pisa` back to your own `master` branch.

atrettin:
Hi Justin,

yes, in the future I will make a new branch for developing features other than the master branch. In this case, I did intend the changes to show up here, since they are necessary to get the stage working, even though it doesn't strictly have anything to do with the memory leak.

jllanfranchi:
Okay great. No worries, just need to know whether to wait to merge beyond your current commits, and also want to help make development easier in the future.

pisa/stages/data/toy_event_generator.py
```diff
@@ -48,8 +48,8 @@ def __init__(self,
         input_apply_keys = ('initial_weights',
                            'weights',
                            'weighted_aeff',
-                           'nominal_nu_flux',
-                           'nominal_nubar_flux',
+                           'nu_flux_nominal',
+                           'nubar_flux_nominal',
```
jllanfranchi: Why did you change these names? (This will break code in the fridge if you do so...)

jllanfranchi: (I count 38 files with "nominal_nu_flux" in them in the fridge, and only 2 with "nu_flux_nominal"; I am, however, seeing the flux stages in PISA using "nu_flux_nominal" naming convention ... not sure what the story is, but it might be a concern)

atrettin: I changed these names because they were changed in the flux stage of PISA when it was re-named to pi_barr_simple. And yes, it was annoying because it did break my code initially. I'm happy to revert this if we also change back the flux stage.

atrettin: I confirmed that this change happened in the commit https://github.com/IceCubeOpenSource/pisa/commit/a5e66450fc56494689ce858dbff20dc0bbc19765

jllanfranchi: @ts4051 can you justify why you changed the names we're talking about above? Can you fix the code in fridge that change broke, or (what I would probably prefer at this point, but I am open to other opinions...) change back to the original naming scheme and fix whatever that change breaks?

philippeller: Since this re-naming change is (unfortunately) already in some parts of master, gonna merge this PR now

PR #557, opened 2020-01-22
Fix iteration of container dictionaries after Py3 switch.

kayla-leonard:
This should fix issues with how dict.keys() works in Python 3.

### DIFF
--- a/pisa/core/container.py
+++ b/pisa/core/container.py
@@ -11,10 +11,10 @@ from __future__ import absolute_import, print_function
 from collections.abc import Sequence
 from collections import OrderedDict
 
+from itertools import chain
+
 import numpy as np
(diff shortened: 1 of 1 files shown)


pisa/core/container.py
```diff
@@ -20,6 +20,7 @@
 from pisa.core.translation import histogram, lookup, resample
 from pisa.utils.log import logging
 
+from itertools import chain
```
jllanfranchi: This is somewhat nitpicky, but eventually you'll find it useful for reading Python code: Start with `__future__` imports, then group imports from python builtins and the python standard library (`collections`, `itertools`, etc.), then imports from external libraries (`numpy`, etc.), then imports from the local package (`pisa`).

PR #560, opened 2020-02-07
hyper-hyper surfaces

atrettin:
I added a factory class in `utils/hypersurface.py` that produces a Hypersurface with coefficients calculated from splines at given oscillation parameters. There is also a loading function that loads a set of interpolated hypersurfaces from a JSON file. The structure of the input JSON file is simple and consists basically of paths to hypersurfaces fit at different oscillation parameters. A script producing such a file will be added to the fridge.

The `pi_hypersurfaces` stage was adapted to be able to work with interpolated hypersurfaces as well and de-lintered.

Furthermore, a small utility was added in `utils/matrix.py` to test matrices for positive semi-definiteness (psd) and find the unique psd matrix that is closest to a given matrix as measured by the Fronebius norm.


jllanfranchi:
I haven't read your code in detail, but a quick glance shows you aren't using the Numpy docstring conventions. Since our documentation is auto-generated, it's critical that the docstrings follow this convention.

See here for the developer guide: https://icecube.wisc.edu/~peller/pisa_docs/readmes/general_conventions.html#documentation-conventions

which links to the actual docstring convention here (redirected link in the above): https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

or just look at files in PISA -- especially in the core -- to see examples of docstrings in PISA.

atrettin:
My apologies, I'll change the doc-strings accordingly.

atrettin:
Alright, that should do it for the doc-strings!

pisa/core/param.py
```diff
@@ -839,7 +839,7 @@ def __setattr__(self, attr, val):
             if isinstance(val, Param):
                 assert val.name == attr
                 self._params[idx] = val
-            elif isbarenumeric(val):
+            elif isinstance(val, ureg.Quantity):
```
jllanfranchi: Why did you change this? (Just making sure you are fixing a bug, and not introducing a bug for something you haven't considered.)

pisa/utils/matrix.py
```diff
@@ -0,0 +1,144 @@
...
+import scipy.linalg as lin
+
+__all__ = [
+    'is_psd',
+    'fronebius_nearest_psd',
+]
+
+__author__ = 'A. Trettin'
+
+__license__ = '''Copyright (c) 2014-2017, The IceCube Collaboration
```
jllanfranchi: since you presumably just created this file, just list the year 2020 here.

pisa/core/param.py
atrettin: This is a bug fix. The values of Param objects are quantities with units, rather than pure numeric values.

pisa/utils/matrix.py
atrettin: ok, I fixed that!

atrettin:
One more thing that I will add: storage of spline keyword arguments inside the JSON file to make fits reproducible, i.e. someone using the same file should get the same interpolated hypersurfaces. I also want to add support for a piecewise linear mode as an alternative to splines.

pisa/core/param.py
jllanfranchi: I just checked out your code and ran the unit tests. This appears to break the use case (which previously worked) of:

```my_param_set.x = 1```

although I agree it now allows

```my_param_set.x = ureg.Quantity("1 meter")```

Note that there are a lot of unit tests, especially in the core modules. Simply run the file to run the unit tests, and you'll see that you broke one here.

I fixed the bug you have introduced, added a unit test to ensure that the second use case you introduced gets tested from now on, and I will submit a PR against _your_ branch of the code. If the changes I made are okay, then you can merge my PR into your branch and the changes should show up here.

atrettin: Ah ok, thank you!

ts4051:
Please add `imunuit` to the PISA depencies in `setup.py`

atrettin:
Wasn't that done in https://github.com/IceCubeOpenSource/pisa/pull/560/commits/d06a81447647bc97a4087d705e7e8526e717f532 ? Or is there another place missing?

ts4051:
Sorry my bad, checked for it in the files changed and saw it wasn't there. Just realised that this is a second PR, e.g. you already did that in the first PR already I guess.

ts4051:
Went through the code, looks good to me. I tested it with GRECO and found some minor issues, which I've fixed and made a PR to the branch (https://github.com/atrettin/pisa/pull/2).

From my side I'm happy to merge this PR to master once my fixes are integrated.

philippeller:
All comments from @jllanfranchi and @ts4051 seem resolved, so gonna merge this

PR #561, opened 2020-02-24
new tool to retrieve bin index of a SmartArray being histogrammed

bourdeet:
Hi ! 

I've added a set of functions that can allow one to retrieve the bin index in which a particular event falls into, once it is histogrammed with some MultiDimBinning object. This can be useful to track for example the distribution of simulation weights inside each analysis bin.

The bulk of the function architecture comes from a modified version of the functions available in translation.py


pisa/core/bin_indexing.py
```diff
@@ -0,0 +1,243 @@
...
+import numpy as np
+from numba import guvectorize, SmartArray, cuda
+
+from pisa import FTYPE, TARGET
+from pisa.core.binning import OneDimBinning, MultiDimBinning
+from pisa.utils.log import logging, set_verbosity
+from pisa.utils.numba_tools import myjit, WHERE
+from pisa.utils import vectorizer
+
+__all__ = [
```
philippeller: should change to contain actual function of this file

pisa/core/bin_indexing.py
```diff
@@ -0,0 +1,243 @@
+"""
+Functions to retrieve the bin location of each elements
+of an array, inside a Container, based on its specified 
+output binning.
+
+Etienne Bourbeau
```
philippeller: move this to __author__

Also, please add standard license etc (check how other files look like)

pisa/core/bin_indexing.py
```diff
@@ -0,0 +1,243 @@
...
+    sample : list of SmartArrays
+
+
+    binning : PISA MultiDimBinning
+
+    Returns: for each event the index of the histogram in which it falls into
+
+    Notes
+    -----
+    this is only a 2d method right now
```
philippeller: looks like 1, 2 or 3d to me?

jllanfranchi:
FYI, dunno how much you want to make this a "generally useful tool," but you could without _too_ much work make this work generically by attaching a (lazy-loading) version to OneDimBinning. Each OneDimBinning determines what the nature of binning is, and then you can add a `digitize` method to that class, which is the numpy convention for what you're doing (although I hate that numpy.digitize returns 1 for index 0, etc). That function would take an array and bin it. Then MultiDimBinning.digitize dispatches the OneDimBinning.digitize method of its member binnings on each dimension of the input array. (Other functions already do this in MultiDimBinning.)

Somewhat optimized code for looking up which bin a thing lives in (in numba) has already been written for retro in this module here: https://github.com/IceCubeOpenSource/retro/blob/master/retro/utils/geom.py#L481, you can re-use much of that probably.

bourdeet:
Regarding the identical comment, I did try to work out a way to add an option to return only the index in `translation.py`, but I could not make it work with these `guvectorize` function signatures...

pisa/core/bin_indexing.py
bourdeet: Should it just be the functions I want people to use (ie not the functions that are only called internally?

pisa/core/bin_indexing.py
bourdeet: removed the name, but I don't see the standard license in the other core scripts (maps, container, base_stage...

pisa/core/bin_indexing.py
bourdeet: fixed the description

pisa/core/bin_indexing.py
philippeller: right. I meant before it was containing function that were _not_ in the file (like `histogram`)

philippeller:
But then at least try to import functions from translate.py rather than copy&paste them (like `find_index`). And see if there is other stuff that can be outsourced

pisa/core/bin_indexing.py
```diff
@@ -0,0 +1,217 @@
...
+    
+    binning_1d = MultiDimBinning([binning_x])
+    binning_2d = MultiDimBinning([binning_x, binning_y])
+    binning_3d = MultiDimBinning([binning_x, binning_y, binning_z])
+
+
+    # 1D case: check that each event falls into its predicted bin
+    print('TEST 1D:')
+    print('array in 1D: ',x.get(WHERE),'\nBinning: ',binning_1d.bin_edges)
+    indices = lookup_indices([x], binning_1d)
```
philippeller: can you here do an actual test checking wether the returned values are what's expected? E.g. using some sort of `assert indices == expected_indices`

bourdeet: Hey!

Just added a couple of assert test matching the calculated binning with what one would expect from the binning logic (ie, I manually checked that each value in the initial arrays should fall into so and so bin, in the 1D, 2D and 3D case).

while doing so, Justin found a bug in `find_index`: (in translation.py)

The code had :
`if (x < bin_edges[i+1]) or (x <= bin_edges[-1] and i == len(bin_edges) - 1)`

but if `i== len(bin_edges)`, that  first test fails. Python evaluates the first condition and short-circuits if the conclusion is clear. If not, it tests subsequent conditions. So the following does work in regular python:

`if (x <= bin_edges[-1] and i == len(bin_edges) - 1) or x < bin_edges[i+1]`

bourdeet:
I agree that I could fix `find_index`, but the changes I make there will have an impact on the other binning/array conversion tools. So before I do that, I'd like to make sure we use the same convention everywhere.

This leads me to the question: what is defined as the underflow and overflow bins in pisa? Is it in fact -1 for the underflow and Nbin for the overflow, like in my code?

philippeller:
Hey, I agree it's a good idea to move the logic of checking whether a value is inside the binning be moved into `find_index`, which will simplify code in several functions. These functions then just proceed depending on the return of `find_index`.
Will you make these changes?

jllanfranchi:
@bourdeet: We don't have (a separate concept of) underflow or overflow bins in PISA. The analyses we have done simply ignore values outside of the defined binning. (Though I understand why one would want such bins as a separate check, it would be tricky to implement at this point.)

If you do want to achieve the same thing as underflow/overflow bins with current code unmodified, you can define a binning like `[-inf, -2, -1, 0, 1, 2, inf]` and that should work just fine (and should be unit-tested that it works correctly for these functions). Of course if you treat `[-inf, -2]` and `[2, inf]` will be used in the analysis, I forget if there's a simple way to define that binning yet slice them out for subsequent analysis.

This sounds like something that @philippeller has thought about, though, already, as he has worked on arbitrary translation among {samples, histograms}: samples -> hist, hist-> samples, samples -> different samples, hist -> different hist. Just not sure how much of that has made its way back into PISA yet.

jllanfranchi:
@bourdeet: We don't have (a separate concept of) underflow or overflow bins in PISA. The analyses we have done simply ignore values outside of the defined binning. (Though I understand why one would want such bins as a separate check, it would be tricky to implement at this point.)

If you do want to achieve the same thing as underflow/overflow bins with current code unmodified, you can define a binning like `[-inf, -2, -1, 0, 1, 2, inf]` and that should work just fine (and should be unit-tested that it works correctly for these functions). Of course if you treat `[-inf, -2]` and `[2, inf]` will be used in the analysis, I forget if there's a simple way to define that binning yet slice them out for subsequent analysis.

This sounds like something that @philippeller has thought about, though, already, as he has worked on arbitrary translation among {samples, histograms}: samples -> hist, hist-> samples, samples -> different samples, hist -> different hist. Just not sure how much of that has made its way back into PISA yet.

bourdeet:
I've made the changes to find_index so that it returns what I mentionned before (-1 for underflow, nbins for overflow). I removed the superfluous conditions in the lookup functions, but I could not do the same in the `tranlsation.py` functions. Somehow I get segfaults, and since the original functions work well I won't change these any further

philippeller:
Ok, thanks for looking into it. Also don;t know why i seg faults, but from my side we can merge now

bourdeet:
so... can we merge?

jllanfranchi:
Please replace all print statements with logging.{warn, info, error} statements.

Would be nice if you could run your new file you introduce into `pisa.core` through the `black` auto-formatter to clean it up before submitting, but I'm fine merging even if you don't share that level of obsessive-compulsiveness.

More importantly, you should run `pylint` over it to check for errors. I already see a bunch of errors and warnings popping up when I just checked this code out.

jllanfranchi:
I'm running it quickly and cleaning up. Will submit a PR against your branch in a minute.

pisa/core/translation.py
```diff
@@ -361,18 +361,31 @@ def find_index(x, bin_edges):
...
-            if (x < bin_edges[i+1]) or (x <= bin_edges[-1] and i == len(bin_edges) - 1):
-                break
+    
+    #
+    # First check: extreme bin
+    #
+    if x<bin_edges[0]:
+        return -1
+    elif x>bin_edges[-1]:
+        return len(bin_edges)-1
```
jllanfranchi: shouldn't this be `return len(bin_edges)`?

bourdeet: the overflow bin has a value of nbin, which has one unit less than the length of the bin edges

philippeller: if there are N bins, we have N+1 bin edges. The last bin has index N-1, so the overflow should be N = `len(bin_edges) - 1`, if i did my math correctly?

philippeller: ah ok, we were typing answers at the same time ;)

jllanfranchi: Very well, my mistake. I'm adding some variables for clarity in my PR.

bourdeet:
Just to make sure @jllanfranchi : are you running pylint and fixing the print statements or should I?

jllanfranchi:
I'm doing that. PR will be submitted in 30 sec

jllanfranchi:
Is it intentional that the bin indices found should be the same type as the input array, or should it be some integer type?

bourdeet:
Sure! If it relates to the expected signature of the function, I just copy-pasted the signatures from previous items. It could very well be that I introduced a frivolous requirement regarding the integer precision

bourdeet:
it was intentional to return integers though, since array indices are using only accepted if they are integers

jllanfranchi:
Makes sense, you copied the signature from the histogramming functions which, due to the possibility that weights are present, returns `f4` (i.e., 4 byte floats, aka float32) for `f4` inputs and returns `f8` for `f8` inputs. But for looking up bin indices, I made all return types `i8` (8 byte signed integers).

By the way, fun fact: you can multiply `OneDimBinning` objects to obtain a `MultiDimBinning` object.

```python
x = OneDimBinning(...)
y = OneDimBinning(...)
binning_2d = x * y
```

PR #562, opened 2020-02-25
Data pipeline mod

bourdeet:
Adding the possibility for a data pipeline to load the livetime directly from the metadata of the hdf5 that gets loaded


pisa/core/events_pi.py
```diff
@@ -163,7 +167,12 @@ def load_events_file(self, events_file, variable_mapping=None):
                     )
 
         if isinstance(events_file, str):
-            input_data = from_file(events_file)
+            input_data,metadata = from_file(events_file,return_attrs=True)
```
philippeller: is this metedata always present? Asking for backwards compatibility

jllanfranchi:
* Isn't it true that if `livetime` is a param required by this stage, then the user has to specify it in their config file? This has a couple of issues:
  * This will overwrite the value specified in the config file, and therefore the user can think one thing is happening but be unaware that something else is happening. If I'm correct about that, then it seems to me that this should just be a double-check vs. an existing `livetime` param, not setting it.
  * Even if all that is successful, you are changing what the expected params are for this stage dynamically at run-time, depending on the input. That's of course something you can do, because you can do anything. And we have probably violated this adage a million times in PISA, but it's extra mental load on users to have to be aware of a different API for a service depending on its input

Therefore, I propose you simply use this as a check: If it's detect-able that the input data file is detector data, and that data file contains a "livetime" in its metadata (is it possible for the file include its units, btw?), **and** "livetime" is an existing param, then make sure that `param.livetime`:

* is equal to the livetime metadata from the file 
* is set to be fixed (it should never be modified by a minimizer!)

philippeller:
I think it is a good idea that it performs a check instead of setting the value! This way it can produce a warning if these don;t match, but one can think of scenarios where one would like to intentionally set a different livetime (test fitting convergence, or generate different pseudo-data, etc etc)

pisa/core/events_pi.py
bourdeet: that metadata is always present, at least based on how pisa hdf5 files are created

jllanfranchi:
People don't pay attention to warning messages. Therefore, the default behavior should be to error out (raise an exception) if the value in the file doesn't match an existing `params.livetime`.

To enable the case of dynamically (or statically, in the config file) modifying `params.livetime` to disagree with `livetime` in the source file, the `simple_data_loader` service should have an instantiation arg (i.e., an argument to `__init__`) that's called something like `allow_livetime_mismatch=False` that defaults to False but can be specified in a config file to be `True` and, in this case, the agreement check (and check for static-ness of `livetime`) is either bypassed or a simple warning is issued.

philippeller:
Ok, maybe i'm confused here. What exact `livetime` are we talking about here?
If you load data, there is no parameter, and if you load MC, you set `lifetime` to whatever your heart desires, no?

bourdeet:
If I load data, I want to extract the livetime of the data that has been saved into that particular file. It could become handy if, for example, we want to handle multiple seasons separately

bourdeet:
The issue is that you don't want to have to double check that the livetime you enter in the config file is the same as the one you take from the files that have been harvested. Computing the total livetime of the hdf5 file is a lengthy and annoying process, that should be handled during harvesting.

ts4051:
The desired behaviour here is that if the detector data file specifies its own livetime this is used and if not (either because you're using MC-based pseudodata, or for backwards compatibility with older PISA detector data files that don't contain the livetime) then use the value from the config file.

The suggested check that the livetime is fixed when using the data file is a good idea. I don't agree with enforcing that the livetime MC pipeline(s) is identical though, I think overwriting (with a warning) is a better solution as it easily allows you to load various different data subsets (e.g. compare different seasons) without needing to make lots of virtually identical copy-pasted MC pipeline cfg files.

philippeller:
Hmm, ok. So what you're porposing, if i understand correctly, is that the data pipeline sets the parameter of the MC pipeline?
That is at least a bit _unorthodox_ in terms of how PISA works under the hood, and not sure whether or not easily doable

jllanfranchi:
I think this logic/functionality might make sense if it were contained in a higher-level object / code. E.g., `DistributionMaker` or the code running the analysis. It violates the boundaries of the objects that one `Pipeline` tries to magically modify the params of another `Pipeline` (that presumably happen to be contained by the same `DistributionMaker`). Since the `DistributionMaker` sees all `Pipelines` that live within it, though, it makes sense for the `DistributionMaker` to perform a check like this.

I'll rough out a simple (possibly not working, I'm not worrying about details of the hdf5 files right now) version of this, submit an alternative PR, and we can discuss which implementation makes more sense (and see if what I write will achieve what you want).

bourdeet:
That sounds like a good idea. It does make more sense to implement it at a level where one is aware of multiple pipelines. 
I think the hdf5 file handling I got should work in both cases

bourdeet:
After fixing a few remaining bugs from PR # 563, I've merged Justin's changes into my own PR.

If I did things correctly this would mean that this PR has now the most up-to-date version of the detector data livetime handling

bourdeet:
Alright! I think this PR is ready to merge, once @philippeller  or @jllanfranchi give the okay

Note that I deactivated the attribution of units to the data_derived livetime, as it was conflicting with the units that one would get by grabbing the livetime from the config file of a stage.

pisa/core/distribution_maker.py
```diff
@@ -87,17 +99,75 @@ def __init__(self, pipelines, label=None):
...
+                                stage_idx,
+                                stage.metadata["livetime"],
+                                data_run_livetime,
+                            )
+                        )
+            #
+            # Set param `params.livetime` for any pipelines that have it
+            #
+            if data_run_livetime is not None:
+               # data_run_livetime *= ureg.sec
```
jllanfranchi: Why did you remove units here? Have you figured out how to handle units elsewhere? (This would be preferred to always using seconds, so removing this should eventually be the goal, but I expect it will take a little more work to make it work...)

pisa/stages/data/simple_data_loader.py
```diff
@@ -78,7 +78,7 @@ def __init__(self,
         # instead of adding params here, consider making them instantiation
         # args so nothing external will inadvertently try to change
         # their values
-        expected_params = ()
+        expected_params = ('livetime')
```
jllanfranchi: "livetime" shouldn't be listed in this stage as an expected param

pisa/core/distribution_maker.py
```diff
@@ -87,17 +99,75 @@ def __init__(self, pipelines, label=None):
...
+
+                    if "livetime" not in pipeline.params.names:
+                        continue
+
+                    pipeline.params.livetime.is_fixed = True
+
+                    if pipeline.params.livetime != data_run_livetime:
+
+                        logging.warning(
+                            "Pipeline index %d has params.livetime = %f, in"
```
jllanfranchi: change `%f` to `%s` here and in the next instance as well...

pisa/core/distribution_maker.py
```diff
@@ -87,17 +99,75 @@ def __init__(self, pipelines, label=None):
...
+                            " disagreement with data livetime = %f defined by"
+                            " data. The pipeline's livetime param will be"
+                            " reset to the latter value and set to be fixed"
+                            " (if it is not alredy).",
+                            pipeline_idx,
+                            float(pipeline.params.livetime.value),
```
jllanfranchi: ... and then don't convert to float. It is always useful to display units with the number in case they are disparate.

You could, e.g., have specified `1 common_year` for one and `1 year` for the other; easy mistake to make, but `float()` of each will yield the same value in both cases, rendering the error message unhelpful.

You are right to use `.value` attr, though, that was a bug on my behalf.

pisa/core/distribution_maker.py
bourdeet: As noted above, it caused issues when comparing the livetime obtained from the files with the livetime obtained from the config file (since the latter are defined so far as dimensionless). We could very well put it back in, if it's possible to define units at the config file level

pisa/stages/data/simple_data_loader.py
bourdeet: If I don't list it there, the stage throws an error saying there are too many params. The only way I found to fix this was to define the livetime param in all pipelines (setting it to -1 in simulation), and make it a requirement of the simple_data_loader

pisa/core/distribution_maker.py
bourdeet: I used the float conversion because otherwise the warning message wouldn't run properly. Will remove now

pisa/core/distribution_maker.py
bourdeet: ok, the units are now back

jllanfranchi:
@philippeller if you're cool with this now, then you or I can merge asap

philippeller:
Hey! Looks ok to me. We can get old behaviour with setting livetime_from_data to false, correct?
Will merge if you can just quickly confirm this

bourdeet:
Hi @philippeller , that is correct, setting livetime_from_data to false will revert to the normal behaviour

PR #565, opened 2020-03-06
New option to skip a service, stage during pipeline config parsing

bourdeet:
The key change is in the way `config_parser.py` reads the `order` list object containing the services and stages to run. If the user passes a list of tuples to `skip_stages`, these pairs of services/modules will be removed from the original list, and the rest of the pipeline parsing will proceed as usual.

The original order of the pipeline stages should be preserved  in this method.

Note that this option assumes that you can only call one stage per service. I'm not sure if there are cases when you would expect several stages to be called from the same service.

Note! This PR also includes a slight bug fix on my previous PR #561 , where I was missing `pisa.core` in my import of the translation.py module.


pisa/utils/config_parser.py
```diff
@@ -528,13 +528,16 @@ def parse_param(config, section, selector, fullname, pname, value):
     return param
 
 
-def parse_pipeline_config(config):
+def parse_pipeline_config(config,skip_stages=None):
```
jllanfranchi: space after comma

pisa/utils/config_parser.py
```diff
@@ -528,13 +528,16 @@ def parse_param(config, section, selector, fullname, pname, value):
...
     """Parse pipeline config.
 
     Parameters
     ----------
     config : string or ConfigParser
 
+    skip_stages: None or List of Tuples identifying
+                 (service, stage) to skip in the parsing
```
jllanfranchi: only 4 spaces additional indent, don't line up with the colon above (see the developer's guide / Numpy docstring convention for any questions about docstring conventions in PISA)

pisa/core/pipeline.py
```diff
@@ -84,9 +84,9 @@ class Pipeline(object):
 
     """
 
-    def __init__(self, config):
+    def __init__(self, config, skip_stages=None):
```
jllanfranchi: The terminology in PISA is "stage" and "service". A "service" is the implementation of a "stage". E.g. `flux` is a stage; one "service" that provides flux would be `pi_mceq_barr`. For some stages, you could have more than one service, and it's possible to specify multiple stages (implemented by different services) in the same pipeline. I.e., `skip_services` is a more correct way to say this. And to specify a particular service, you have to provide `(stage, service)`.

Also, add documentation to the above "Parameters" section in the class docstring for this new argument.

pisa/utils/config_parser.py
```diff
@@ -614,6 +617,26 @@ def parse_pipeline_config(config):
...
+
+    #
+    # Browse through the list of (service, stage) Tuple to skip, if any
+    #
+    if skip_stages is not None:
+        # Check the types of skip_Stages. must be a list of Tuples
+        assert isinstance(skip_stages,list) and isinstance(skip_stages[0],tuple), 'ERROR: skip_stages must be a list of Tuples'
+
+        # reformat order into a list of tuples
+        order = [(x[0],x[1]) for x in order]
```
jllanfranchi: space after comma; all you have to do here is:
```python
order = [(x[0], x[1]) for x in order if (x[0], x[1]) not in skip_services]
```

pisa/utils/config_parser.py
```diff
@@ -528,13 +528,16 @@ def parse_param(config, section, selector, fullname, pname, value):
...
 
-def parse_pipeline_config(config):
+def parse_pipeline_config(config,skip_stages=None):
     """Parse pipeline config.
 
     Parameters
     ----------
     config : string or ConfigParser
 
+    skip_stages: None or List of Tuples identifying
```
jllanfranchi: one space before colon; following is more precise type specification:
`sequence of 2 strings, sequence of these, or None; optional`

pisa/utils/config_parser.py
```diff
@@ -614,6 +617,26 @@ def parse_pipeline_config(config):
     # Get and parse the order of the stages (and which services implement them)
     order = [split(x, STAGE_SEP) for x in split(config.get(section, 'order'))]
 
+
+    #
+    # Browse through the list of (service, stage) Tuple to skip, if any
+    #
+    if skip_stages is not None:
+        # Check the types of skip_Stages. must be a list of Tuples
+        assert isinstance(skip_stages,list) and isinstance(skip_stages[0],tuple), 'ERROR: skip_stages must be a list of Tuples'
```
jllanfranchi: Assertion errors are less useful than raising a TypeError in the case that types are incorrect for other people's code and for unit testing. But input shouldn't need to be a list, just the more generic abstract base class (abc) `Sequence`. And why do the inner sequences have to be tuples? Why should a user expect a list of two items specifying `[stage, service]` to be illegal?

Also, does a user specifying a single (stage, service) have to pass that as [(stage, service)]? Why not simplify the user's life, and allow `(stage, service)` or `[(stage, service)]` or `[(stage, service), (stage, service), ...]`? Code below should implement this in a fairly concise manner


```python
# at the top of the file in the "python builtin imports" section
from collections.abc import Sequence


# ... inside your function, preferably at the top to

if not (skip_services is None or isinstance(skip_services, Sequence)):
    raise TypeError(
        "`skip_services` expected to be None or Sequence, got {}".format(type(skip_services))
    )

if skip_services is None:
    skip_services = []
else:
    if isinstance(skip_services[0], str):
        skip_services = [skip_services]

    normalized_skip_services = []

    for skip_service in skip_services:
        if not isinstance(skip_service, Sequence):
            raise TypeError("expected Sequence, got {}".format(type(skip_service))

        if len(skip_service) != 2:
            raise TypeError(
                '(stage, service) sequences be length 2; got "{}", of length {}'.format(
                    skip_service, len(skip_service)
                )
            )

        for item, name in zip(skip_service, ["Stage", "Service"]:
            if not isinstance(item, str):
                raise TypeError(
                    "{} must be specified as a string; got {}".format(name, type(item))
                )

        normalized_skip_services.append((skip_service[0], skip_service[1]))

    skip_services = normalized_skip_services
```

bourdeet: the restriction to tuples was made so that the pairs could be easily compared within the two lists (I initially used sets type to find the difference between two sets of tuples, but it didn't preserve the order). You're right that with the way things are implemented now, the type of the skip_service can be a bit more flexible.

Adding various type options like this sounds good.

pisa/utils/config_parser.py
bourdeet: I would rather write:

`order = [[x[0],x[1]] for x in order if (x[0],x[1]) not in skip_services]` so that `order` keeps its original format of a list of list

jllanfranchi: That's fine, but then there's no reason (and in fact it's inconsistent) to construct the list of tuples up above (as I had done). So make `skip_services` into a list of 2-lists as well anywhere that is done, and I'm fine with this. (but space after a comma)

bourdeet:
I have to make data-mc plots using the files we are passing through the pipelines we use for data analysis. I must produce plots with  the nominal and the best fit point. This means I have to have access the array data, which means I have to get them before events get histogrammed. I also want the ability to turn on and off certain services in the pipeline, because the goal is precisely to check the effect of the analysis chain on the output weights. Making tons of different pipeline config files will inevitably lead me to copy paste the same slightly different calls to stages, and none of them will reflect what actually goes on in the analysis.

jllanfranchi:
I understand you don't want to make multiple config files. But do you use a Python script to perform your plotting, which calls DistributionMaker or Pipeline or whatever with a config file?

bourdeet:
Yes, my script calls in Distribution Maker and I feed whatever comes out of `get_outputs` to the plotting tools

jllanfranchi:
Okay, then can you have your script implement the logic you want? E.g. something like the following (where you want to leave out one service):

```python

orig = parse_pipeline_config(filename)
new = OrderedDict()
for key, stage_dict in orig.items():
    if key not in skip_stages:
        new[key] = stage_dict

orig_dm = DistributionMaker(orig)
new_dm = DistributionMaker(new)
...
```

PR #567, opened 2020-03-11
Merging in generalised NSI implementation from Thomas Ehrhardts branch

ellohfin:
Merging in generalised NSI implementation from Thomas Ehrhardts thehrh/pisa nsi_reparameterisation branch. 
Comparisons were made for Oscillograms and Asimov-scans of eps_tautau - eps_mumu and eps_ee - eps_mumu, comparing to what is shown in Thomas' NSI analysis wiki (oscillograms as in section Oscillation probability calculation; scans as in 1D studies in the NSI sensitivities section). This yielded no differences.


pisa/stages/osc/nsi_params.py
```diff
@@ -2,12 +2,34 @@
...
+            'Pass an iterable with two items (magnitude and phase)!'
+        )
+    if magnitude < 0.0 and phase != 0.0:
+        raise ValueError(
+            'Only accepting negative values with a zero phase (real coupling)!'
+        )
+    return magnitude, phase
+
+
 class NSIParams(object):
```
thehrh: This base class seems to be an outdated version compared to https://github.com/thehrh/pisa/blob/nsi_reparameterisation/pisa/stages/osc/nsi_params.py. Can you explain its goal?

pisa/stages/osc/nsi_params.py
```diff
@@ -115,3 +137,560 @@ def eps_tautau(self, value):
...
+    def eps_ee(self):
+        """effective nue-nue NSI coupling parameter"""
+        return self.eps_matrix[0, 0].real
+
+    @eps_ee.setter
+    def eps_ee(self, value):
+        part1 = self.eps_matrix[0, 0].imag
+        part2 = 1.j * part1
+        part3 = value + part2
+        self.eps_matrix[0, 0] = part3
```
thehrh: Is this duplication of setting the matrix element intended?

pisa/stages/osc/nsi_params.py
ellohfin: I'm afraid I don't quite get or see what you mean

ellohfin: oh, sorry, now I got it I think

pisa/stages/osc/nsi_params.py
ellohfin: no, I'll remove the less verbose version

pisa/stages/osc/pi_osc_params.py
```diff
@@ -3,6 +3,10 @@
 """
 OscParams: Characterize neutrino oscillation parameters
            (mixing angles, Dirac-type CP-violating phase, mass splittings)
+
+changed by Elisa Lohfink (ellohfin; elohfink@icecube.wisc.edu)
```
thehrh: You seem to have left the NSI coupling getters and setters in this class. Probably not on purpose?

thehrh: There's also some (empty) dummy unit testing functionality at the bottom of https://github.com/thehrh/pisa/blob/nsi_reparameterisation/pisa/stages/osc/pi_osc_params.py, which could be useful to keep so the `TODO` is at least documented.

pisa/stages/osc/pi_prob3.py
```diff
@@ -108,11 +170,35 @@ def __init__(self,
...
+            )
+        if self.nsi_type == 'vacuum-like':
+            logging.debug('Working in vacuum-like NSI parameterization.')
+            self.nsi_params = VacuumLikeNSIParams()
+        elif self.nsi_type == 'standard':
+            logging.debug('Working in standard NSI parameterization.')
+            self.nsi_params = StdNSIParams()
 
         # setup the layers
         #if self.params.earth_model.value is not None:
```
thehrh: If you compare to my branch, in ll. 206-208 I set `self.YeI` etc., and in 212 pass `self.YeI` etc... Without those changes, the check for changed electron fractions in l. 283 won't work.

pisa/stages/osc/pi_prob3.py
```diff
@@ -186,14 +276,67 @@ def compute_function(self):
             self.data.link_containers('nubar', ['nuebar_cc', 'numubar_cc', 'nutaubar_cc',
                                                 'nuebar_nc', 'numubar_nc', 'nutaubar_nc'])
 
+        # this can be done in a more clever way (don't have to recalculate all paths)
+        YeI = self.params.YeI.value.m_as('dimensionless')
+        YeO = self.params.YeO.value.m_as('dimensionless')
+        YeM = self.params.YeM.value.m_as('dimensionless')
+        if YeI != self.YeI or YeO != self.YeO or YeM != self.YeM:
```
thehrh: cf. above

pisa/stages/osc/pi_prob3.py
thehrh: cf. https://github.com/thehrh/pisa/blob/nsi_reparameterisation/pisa/stages/osc/pi_prob3.py#L206

thehrh:
The file `numba_osc.py` is not that from my branch. It seems like the only difference (beside a few comments) is that `mat_pot` is still named `eps_mat`, which I'm almost certaint will lead to confusion :sweat_smile: . Therefore, I'd encourage you to go with my naming convention.

thehrh:
I believe what I've commented on are all the relevant deviations from the same four files in my branch.

pisa/stages/osc/nsi_params.py
ellohfin: Do you mean the part where I stuck to the PISA master way of doing this instead of the NSI branch way? Or do I misunderstand what you mean?

thehrh: Yes. The functionality in this version of the base class is meant to be superseded by the two derived classes, in particular by `StdNSIParams`. You can also compare the docstrings of `NSIParams` here and in https://github.com/thehrh/pisa/blob/nsi_reparameterisation/pisa/stages/osc/nsi_params.py.

Unless you have any particular usage of this (in my eyes outdated) version of `NSIParams` in mind?

pisa/stages/osc/pi_prob3.py
ellohfin: Assuming the "if" is intended to do what it does, I understand this should stay the way it is, just like in your branch?

thehrh: Right

jllanfranchi:
@ellohfin @thehrh how far out is this from being able to be merged?

philippeller:
There is one merge conflict, can someone `git merge master` into that PR branch and resolve?

thehrh:
Were you planning on fixing the merge conflicts and performing a few more tests on your latest commit?

ellohfin:
yes, cobalt was down so I didn't do that directly

pisa/stages/osc/pi_osc_params.py
```diff
@@ -62,10 +80,6 @@ def __init__(self):
         self.dm21 = 0.
         self.dm31 = 0.
         self.dm41 = 0.
-        self.nsi_eps = np.zeros((3, 3), dtype=FTYPE) + 1.j * np.zeros((3,3), dtype=FTYPE)
-        self.gamma21 = 0. # TODO Add full 3x3 matrix option, TODO update docs, TODO getters/setters to enforce values ranges?
```
thehrh: I doubt you want to remove the parameters `gamma_ij` here

pisa/stages/osc/nsi_params.py
```diff
@@ -1,117 +1,615 @@
...
+        self.eps_matrix[2, 2] = value + 1.j * self.eps_matrix[2, 2].imag
+
+class VacuumLikeNSIParams(NSIParams):
+    """
+    NSI parameters using a vacuum Hamiltonian-like parameterization.
+
+    """
+
+    def __init__(self):
+        super(NSIParams, self).__init__()
```
thehrh: `super(VacuumLikeNSIParams, self).__init__()`

ellohfin:
...I'll now try try run the tests locally, first.

ellohfin:
The tests of nsi_params.py fail because sympy isn't present. We can
1) test this locally and ignore the test failing here or 
2) put sympy in the test environment.
Since sympy is only needed for testing nsi_paramy.py I'd go with 1) for now. Any opinions on that?

pisa/stages/osc/prob3numba/numba_osc.py
```diff
@@ -879,3 +912,5 @@ def fill_probs(probability, initial_flav, flav, out):
     test_convert_from_mass_eigenstate()
     test_get_transition_matrix()
     test_osc_probs_layers_kernel()
+
```
jllanfranchi: get rid of extra blank lines at end of file

pisa/stages/osc/prob3numba/numba_osc.py
```diff
@@ -115,27 +120,30 @@ def get_H_mat(rho, nsi_eps, nubar, H_mat):
     # 2*sqrt(2)*Gfermi in (eV^2-cm^3)/(mole-GeV)
     tworttwoGf = 1.52588e-4
     a = 0.5 * rho * tworttwoGf
-    if nubar == -1:
-        a = -a
 
     # standard matter interaction Hamiltonian
     clear_matrix(H_mat)
-    H_mat[0,0] = a
+    H_mat[0, 0] = 0
```
jllanfranchi: I believe `clear_matrix()` sets all elements of a matrix to 0, so this should be unnecessary, I think...

pisa/stages/osc/prob3numba/numba_osc.py
```diff
@@ -89,16 +89,21 @@ def test_get_H_vac():
...
-        density
+        electron number density (in moles/cm^3)
+        (numerically, this is just the product of
+        electron fraction and mass density in g/cm^3,
+        since the number of grams per cm^3 corresponds to
+        the number of moles of nucleons per cm^3)
 
-    nsi_eps : complex 2-d array
-        Non-standard interaction terms
+    mat_pot : complex 2-d array
```
jllanfranchi: please include in docstring that one should make this `np.diag([1, 0, 0])` for standard interactions only (here and anywhere else this kwarg is used). Also, make the docstrings consistent for this wherever they appear.

pisa/stages/osc/pi_osc_params.py
jllanfranchi: (Note that raising `NotImplementedError` will keep our whole-PISA-unit-test-runner-script from passing, so preferably that is turned into at least some tests, but if not, that should be just turned into a `pass` statement for now and e.g. add a `# TODO` comment or something.)

ellohfin: Can't such an error be handled while testing?

thehrh: My bad. I know it prevents it from passing. I found that awareness of a lacking unit test should be escalated.

jllanfranchi: @ellohfin : yes, we could add "allow the following errors to pass but turn them into warnings" or something logic. But that's a lot more work than I'm willing to put into it, and then we end up with a lot of weird stuff in the unit test runner script, and this kind of thing just gets ignored ayway. You could achieve a similar result by emitting a `logging.warning` message. Or... add some actual unit tests, rather than make me have to code around the lack of unit tests!

@thehrh : I get it, I've done this in the past, too, to make it more visible. We could add this in about a thousand places throughout PISA, ... :(

Overall, my philosophy is: We need lots more unit testing in PISA. We need people to feel unhappy about their own code until they have written good unit tests for it. BUT we also want Github CI tool to help us out, and it only helps us if a red X means someone broke something vs a green checkmark means they haven't broken anything (that was tested to work previously).

Any behavior besides that:, anything with any more subtlety than X vs Check, any requirements to read output, and the Github CI tool is useless (or not _useless_ per se, but it just will be _ignored_ by nearly everyone).

thehrh: Ok, I'm convinced. The side effects of purposely breaking the testing script would be pretty severe (given our current unit-testing status).

PR #568, opened 2020-03-12
Bugfix: find_index was incompatible with numpy.histogramdd

jllanfranchi:
`find_index` and `test_find_index` updated to treat bin edges as follows:

```
[ bin 0 ) [ bin 1 ) ... [ bin num_bins - 1 ]
```

 Also, handle nan: If value passed is `nan`, return -1 (underflow index). Definition for how to handle things is taken to be defined by behavior of `np.histogramdd` function. Unit test is reworked to use that as definition as well.

Unit test now includes +/-inf, integer and non-integer, positive and negative bin edges, and the binary-un-representable value 0.1. Values tested as inputs include +/-inf, nan, all bin edges, and floating point resolution above and below each bin edge.

### DIFF
--- a/pisa/core/translation.py
+++ b/pisa/core/translation.py
@@ -21,9 +21,14 @@ from pisa.utils.numba_tools import myjit, WHERE
 from pisa.utils import vectorizer
 
 __all__ = [
+    'resample',
+    'get_hist',
     'histogram',
     'lookup',
+    'find_index',
     'resample',
+    'test_histogram',
+    'test_find_index',
 ]
 
 
@@ -299,7 +304,7 @@ def lookup(sample, flat_hist, binning):
 
     """
     #print(binning)
-    assert binning.num_dims in [2,3], 'can only do 2d and 3d at the moment'
+    assert binning.num_dims in [2, 3], 'can only do 2d and 3d at the moment'
     bin_edges = [edges.magnitude for edges in binning.bin_edges]
     # todo: directly return smart array
     if flat_hist.ndim == 1:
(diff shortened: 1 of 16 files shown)


pisa/core/translation.py
```diff
@@ -360,38 +362,56 @@ def lookup(sample, flat_hist, binning):
...
-                    break
-                else:
-                    first = i + 1
+        # Binary search within binning (inclusive of left and right edges)
+        left_idx = 0
+        right_idx = num_bins
+        while left_idx < right_idx:
+            idx = left_idx + ((right_idx - left_idx) >> 1)
+            if val >= bin_edges[idx]:
+                left_idx = idx + 1
```
philippeller: i'm not mentally capable of being 100% sure this is correct, but it looks like. Also I see you tested it.

jllanfranchi: I updated some variable names and added comments in hopes that the logic is a bit easier to follow (binary search still isn't trivial, but should be able to reason it out). And added more unit tests (e.g. for single-bin binnings).

jllanfranchi:
I believe your test function's "reference" value is incorrect, as opposed to `find_index` being incorrect. Here's the output (I added the error message for the AssertionError at the bottom to show what values are being tested and what values are considered "reference")

```
$ PISA_FTYPE=fp32 python ~/src/pisa/pisa/core/bin_indexing.py 
<< PISA_FTYPE env var is defined as: "fp32"; PISA is running in single precision (FP32) mode; numba is running on CPU (single core) >>
[   TRACE] TEST 1D:
[   TRACE] Total number of bins: 7
[   TRACE] array in 1D: [-5.   0.5  1.5  7.   6.5  8.   6.5]
[   TRACE] Binning: 0.0 dimensionless
[   TRACE] indices of each array element: [-1  0  1  6  6  7  6]
[   TRACE] *********************************
[   TRACE] TEST 2D:
[   TRACE] Total number of bins: 28
[   TRACE] array in 2D: [(-5.0, -5.0), (0.5, 0.5), (1.5, 1.5), (7.0, 1.5), (6.5, 3.0), (8.0, 1.5), (6.5, 2.5)]
[   TRACE] Binning: [<Quantity([0. 1. 2. 3. 4. 5. 6. 7.], 'dimensionless')>, <Quantity([0. 1. 2. 3. 4.], 'dimensionless')>]
[   TRACE] indices of each array element: [-1  0  5 25 27 28 26]
[   TRACE] *********************************
Traceback (most recent call last):
  File "/home/justin/src/pisa/pisa/core/bin_indexing.py", line 294, in <module>
    test_lookup_indices()
  File "/home/justin/src/pisa/pisa/core/bin_indexing.py", line 269, in test_lookup_indices
    assert np.array_equal(test, ref), "test={} != ref={}".format(test, ref)
AssertionError: test=[-1  0  5 25 27 28 26] != ref=[-1  0  5 25 26 28 26]
```

bourdeet:
I think you're right. I redid the math and it seems like point (6.5,3.0) should indeed be in bin 27

bourdeet:
Same goes for the 3D case. The correct bin for (6.5,3.0,0.5) should be 54

If you could correct the test in bin_indexing that'd be great!

jllanfranchi:
Thanks for checking, will update the test accordingly

Issue #570, opened 2020-03-15
Fix matrix.py unit test


jllanfranchi:
* Does the unit test check against anything?
* Make unit test into its own function that is named `test_<something>`, and then this function gets called in the `if __name__ == "__main__":` code block (this makes it compatible with `py.test` unit testing package when someday we convert to using that)
* Make PASS output same as other unit tests in PISA: `<< PASS : test_blah >>`; make it exit with an exception if it fails
* Don't print anything else out, we want to read as little as possible when examining the output of a hundred unit tests

atrettin:
Fixed, see pull request #575 .

PR #575, opened 2020-03-17
Fix matrix unit test and make small fix to binning

atrettin:
Should fix issue #570 .

It was possible in `binning` to give an index that is actually invalid. This would cause an obscure error later on when only the length of the now string index is checked. Even more obscure, if the given index string happened to have the same length as the binning has dimensions ("pid"), it would pass the length check and break later with an even more obscure error. 

Let's raise an exception instead with a helpful list of the keys that would have been valid.

### DIFF
--- a/pisa/core/binning.py
+++ b/pisa/core/binning.py
@@ -2730,6 +2730,7 @@ class MultiDimBinning(object):
             for d in self.iterdims():
                 if d.name == index:
                     return d
+            raise ValueError(f"index '{index}' not in {[d.name for d in self.iterdims()]}")
 
         # TODO: implement a "linearization" like np.flatten() to iterate
         # through each bin individually without hassle for the user...
--- a/pisa/core/binning.py
+++ b/pisa/core/binning.py
@@ -2730,7 +2730,7 @@ class MultiDimBinning(object):
             for d in self.iterdims():
                 if d.name == index:
                     return d
-            raise ValueError(f"index '{index}' not in {[d.name for d in self.iterdims()]}")
+            raise ValueError(f"index '{index}' not in {self.names}")
 
         # TODO: implement a "linearization" like np.flatten() to iterate
         # through each bin individually without hassle for the user...
--- a/pisa/utils/matrix.py
+++ b/pisa/utils/matrix.py
@@ -5,6 +5,8 @@ Utilities for performing some not-so-common matrix tasks.
 import numpy as np
 import scipy.linalg as lin
 
+from pisa.utils.log import logging, set_verbosity
+
 __all__ = [
     'is_psd',
     'fronebius_nearest_psd',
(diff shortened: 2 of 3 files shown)


philippeller:
Looks simple enough, will merge today unless someone says otherwise

pisa/core/binning.py
```diff
@@ -2730,6 +2730,7 @@ def __getitem__(self, index):
             for d in self.iterdims():
                 if d.name == index:
                     return d
+            raise ValueError(f"index '{index}' not in {[d.name for d in self.iterdims()]}")
```
jllanfranchi: You can use the `names` attr here rather than having to iterate/list-comprehend yourself:

```python
raise ValueError(f"index '{index}' not in {self.names}")
```

... but this is certainly a useful exception to have for users, so thanks for adding it.

pisa/utils/matrix.py
```diff
@@ -129,16 +129,17 @@ def test_frob_psd(A):
...
-    print(m_test)
-    print('matrix after:')
-    print(fronebius_nearest_psd(m_test))
-    print('The result matrix is psd and the Frobenius norm matches the expectation!')
-    print('testing random matrices...')
     for i in range(100):
         m_test = np.random.randn(3, 3)
         test_frob_psd(m_test)
-    print('Test passed!')
+    sys.stdout.write('<< PASS : test_matrix_random >>\n')
```
jllanfranchi: please import at the top of your file `from pisa.utils.log import logging, set_verbosity` and use `logging.info('<< PASS : test_matrix_random >>')` here, then `set_verbosity(1)` under `if __name__ == '__main__'`.

This allows for central (or fine-grained) control of what kinds of messages PISA emits to the user. Having `print` or `stdout`/`stderr` calls directly takes this away. We don't do a lot to control these messages right now, but having this everywhere might be quite useful e.g. when we get proper py.test integration, and we might simply want _no_ messages emitted (and then we can control that simply).

atrettin: I didn't know how exactly this was supposed to be done, so I copied from the one file where it was done wrongly! I fixed it in `jsons.py` as well now.

jllanfranchi: That was a bad example to take from, unfortunately, as we (at least used to) have circular import issues with `log` importing `jsons` and `jsons` importing `log`. This might be resolved now, but that's why that file in particular was different from the convention we try to follow elsewhere in PISA.

I don't seem to be able to replicate the circular import error with your modifications to jsons.py, so maybe I fixed that issue somewhere along the way and just never realized it... (I'll do a little more testing with your branch here to be sure, though, that it's not a problem any longer).

atrettin:
Sorry for asking a dumb question, but how exactly do I deal with the build fails? I already suspect that it is because the branch I was editing from was not up to date and I'd have to merge the current master branch. Is that the right thing to do? I don't want to mess up this PR...

jllanfranchi:
Don't worry about it for now, Philipp and I can merge this even if there are build failures. Not going to enforce this while we're still transitioning to getting the auto-build / auto-unit testing checks working.

In the future, though, that would be the strategy I'd follow (merge the latest `master` into your feature branch). But there might be some things that my latest (merged) PR makes tricky for you that there's no need for you to have to deal with (this is a "clean" PR right now, it just doesn't pass the tests), so I say just leave this as-is.

PR #576, opened 2020-03-17
Pi kde support

philippeller:
Added a basic KDE stage. Tested with DRAGON datarelease sample.

- This doe not yet contain any support for uncertainties - something we'll need to think about.
- GPU support does not work, as pycuda (what KDE uses) wants exclusive control of the GPU.
- It is _very_ slow for the full neutrino sample, but I guess using this for muons should be fine

### DIFF
--- a/pisa/stages/utils/pi_kde.py
+++ b/pisa/stages/utils/pi_kde.py
@@ -14,8 +14,7 @@ from pisa.utils import vectorizer
 from pisa.utils.kde_hist import kde_histogramdd
 
 class pi_kde(PiStage):
-    """
-    stage to kde map events
+    """stage to KDE-map events
 
     Paramaters
     ----------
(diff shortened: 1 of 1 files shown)


pisa/stages/utils/pi_kde.py
```diff
@@ -0,0 +1,117 @@
...
+from pisa.utils.log import logging
+from pisa.utils.profiler import profile
+from pisa.utils import vectorizer
+from pisa.utils.kde_hist import kde_histogramdd
+
+class pi_kde(PiStage):
+    """
+    stage to kde map events
+
+    Paramaters
```
jllanfranchi: spelling

atrettin:
I'm going to test this on muons now to see if the resulting maps make sense.

atrettin:
These actually look pretty nice! And since the kernel size is adaptive, the higher quality maps do not get over-smoothed. Now, it is a different question if the blob in the lower region of the track bin is real or not, but with an expectation of less than one event in the entire 7 year period I guess those bins just wouldn't matter... 

kde_test_map_muon_pid_cascades
kde_test_map_muon_pid_mixed
kde_test_map_muon_pid_tracks

PR #577, opened 2020-03-17
Pyflake fixes

philippeller:


pisa/stages/utils/pi_kde.py
```diff
@@ -0,0 +1,117 @@
...
+from pisa.utils.log import logging
+from pisa.utils.profiler import profile
+from pisa.utils import vectorizer
+from pisa.utils.kde_hist import kde_histogramdd
+
+class pi_kde(PiStage):
+    """
+    stage to kde map events
+
+    Paramaters
```
jllanfranchi: sp

pisa/stages/utils/pi_kde.py
```diff
@@ -0,0 +1,117 @@
...
+    ----------
+
+    bw_method: string
```
jllanfranchi: put expected params as a list under the documentation of the kwarg `params`

setup.py
```diff
@@ -233,6 +233,7 @@ def do_setup():
             'simplejson>=3.2',
             'tables',
             'uncertainties',
+            'llvmlite<=0.30.0',
```
jllanfranchi: please add a short comment as to why this version limit is imposed, so we are better aware when other things whether or not we can relax this restriction (or if it continues to be necessary)

pisa/stages/utils/pi_kde.py
philippeller: Ah crap, that file is in an separate PR....but now i accidentally committed it here too. Should we do that all in one PR now or want me to remove this file? (can i `git rm`?)

jllanfranchi: Nah, just fix this spelling mistake that other branch, merge that other PR/branch to PISA, merge the  `i3os/master` back into this branch, and then I don't think this file will show up with this PR any longer. Then we can "separate concerns," so-to-speak :wink: ...

PR #581, opened 2020-03-18
allow (all?) scalar types in vectorizer; black

jllanfranchi:
* Create signatures for each known scalar type to avoid explicit casting to one or another in the python code
  * This allows e.g. `pwr` and `min_counts` (in `pow` and `replace_where_counts_gt`, respectively) to be either float, int, or uint, and of any supported width
* Bugfix: order of args to `replace_where_counts_gt_gufunc` was incorrect

### DIFF
--- a/pisa/utils/vectorizer.py
+++ b/pisa/utils/vectorizer.py
@@ -18,24 +18,26 @@ from pisa.utils.numba_tools import WHERE
 
 
 __all__ = [
-    'mul',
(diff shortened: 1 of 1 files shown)


jllanfranchi:
decided to convert all scalars to `FTYPE`; subsequent PR will address the `i4` scalar in `replace_*` and the bug of arg order in that gufunc

PR #587, opened 2020-03-19
Analysis touchups

thehrh:
Finally got around to doing this... I really found no way of splitting this up into smaller chunks, just too  many interwoven changes. There's still a bit of work to do (see bottom), but I wouldn't mind general comments in the meantime

Introducing fit settings (and pull method), new minimizer-settings format, a more easily maintainable main analysis script, collect common routines in central places, generic routine for optimising over discrete hypotheses, more flexible configuration of minimiser starting point, fix some hypo testing shortcomings, ... (see commit logs). There are still some todos (all changes to scripts/analysis.py are still missing, `analysis.nofit_hypo` usage needs to be fixed, probably going to remove the "scan" fit method, and I have to check whether the changes affect analyses outside of pisa).


pisa/analysis/analysis.py
```diff
@@ -265,56 +122,39 @@ def get_separate_t23_octant_params(hypo_maker,inflection_point) :
...
-        (inflection_point,theta23.range[1])
+        (theta23.range[0], inflection_point), (inflection_point, theta23.range[1])
         )
 
     # If theta23 is very close to maximal (e.g. the transition between octants)
     # offset it slightly to be clearly in one octant (note that fit can still
     # move the value back to maximal)
     tolerance = 1. * ureg.degree
-    if np.isclose(theta23.value.m_as("degree"),45.,atol=tolerance.m_as("degree")) : 
+    if np.isclose(theta23.value.m_as("degree"), 45., atol=tolerance.m_as("degree")):
```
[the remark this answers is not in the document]
thehrh: Good find. If this is really necessary, the shift should probably always be away from 45 degree. The routine should also make sure that the shift is compatible with the range, a bug which I fixed for our original octant checking logic (https://github.com/IceCubeOpenSource/pisa/blob/486e64a15096ce636159ddcade3bb2c78f03afdd/pisa/analysis/analysis.py#L450). Here, I overlooked it.

pisa/analysis/analysis.py
```diff
@@ -330,14 +170,412 @@ class Analysis(object):
...
+
+        Parameters
+        ----------
+        #TODO
+
+        """
+        # let someone pass just a single extra param selection
+        # (which could just as well be part of the regular
+        # hypo param selections)
+        if not isinstance(extra_param_selections, Sequence):
```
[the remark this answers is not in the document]
thehrh: Yep, not the first time I this is tripping me up. Will be fixed.

thehrh:
I'm going to work on a notebook that gives a more concise overview of the changes in here (together with context) and post a link here once it exists.

pisa/utils/random_numbers.py
```diff
@@ -82,7 +82,7 @@ def get_random_state(random_state, jumpahead=0):
 
     """
     if random_state is None:
-        new_random_state = np.random
+        new_random_state = np.random.RandomState()
```
jllanfranchi, line 85: FYI, it should be documented, but it was intentional to use Numpy's internal random state if `random_state` is None. We could debate whether this is a smart thing to do, but basically my reasoning was "if you do nothing, you get the behavior as if you just called `np.random.foo`." That is how many people use random numbers in Numpy, whether or not that's a good idea or something we should even allow is up for debate.

philippeller:
Hey, this PR is now already accumulating conflicts. Originally we wanted to wait because the verification sample unblinding was _imminent_. There seem to be some delays however and I wanted to ask if we can go a head and merge anyway?
Otherwise the verification group could make a tagged release

ts4051:
I agree we can't keep waiting here. Is someone able to provide a brief summary of the changes in here that would impact the verification sample analyses? E.g. what has changed in the analysis class, changes to cfg files etc.

We could then decide on whether to just merge and update the other analyses accordingly, or whether we need some tagging/branching as the short term path forwards.

thehrh:
I'll get to it this weekend and fix the conflicts, so that you'll have a summary and a mergeable PR by Monday.

pisa/utils/random_numbers.py
thehrh: Is `np.random.RandomState()` not "Numpy's internal random state"? I.e., I consider the above a fix that's in line with the intention you describe. Calling `get_random_state(get_random_state(None))` (which happens if you try and fluctuate a `MapSet` without passing a seed or a random state) currently fails because of a `TypeError`.

jllanfranchi: No, that's not the case: https://numpy.org/doc/1.16/reference/generated/numpy.random.RandomState.html 
>   If seed is None, then RandomState will try to read data from /dev/urandom (or the Windows analogue) if available or seed from the clock otherwise.

There's an additional wrinkle, in that Numpy's random number generation has changed some since we first wrote PISA, so it's possible that using `np.random` directly will give different behavior from before (i.e., reference a different random number generator, nothing to do with seeding or where its state comes from), whereas the random number generator used in `np.random.RandomState()` has been guaranteed to give identical results in the future as it gave in the past.

Maybe we simply error out in this case. Why is a user doing anything without explicitly supplying a seed in PISA anyway? I can only think of sloppy, bad, and/or wrong things. Thoughts?

thehrh: Ok, good to know. I'm ok with not permitting to pass in `None` here. That would still allow the user to create some random state outside of PISA (if they must).

thehrh: I've decided against touching `get_random_state` (reverted back to its original behaviour for now) because I can't afford the time to look through the code base right now (for example, we have to adapt the docstring of `Map.fluctuate`).

thehrh:
Here's an overview of the changes in my own words, together with what these changes require analysers to do:

https://nbviewer.jupyter.org/urls/seafile.rlp.net/f/c156d1fc084f4bd8af63//%3Fdl%3D1.

I probably missed something, so I wouldn't mind if someone could test whether their (fridge) analysis scripts still work with this branch, after having reformatted their minimizer settings file as detailed above.

philippeller:
Could the current analyzers then give this a try and report possible issues.

bourdeet:
Hey there!

Is there a strong motivation to rename the metadata keys from `minimizer_` to `fit_`? To me this is just introducing unnecessary complications when trying to interpret results file processed before and after this PR

thehrh:
It's a generalisation that's consistent with the naming of our **fit** functions in the `analysis` module. And you might have seen that "minimization" is now not the only fitting method available anymore. I decided against key names depending on the chosen fitting method, as I think a user should be able to read out fit results without knowing which method was used to obtain the result. If people really want the `minimizer_metadata` and `minimizer_time` keys to stay present, I could make identical copies that are available under both names (`fit_` + `minimizer_`) in case minimization was chosen for fitting.

bourdeet:
I'm not sure I get this distinction. How can a fit be done without minimizing something?

thehrh:
Trivial example: you decide to randomly sample your parameter space and look for the maximum likelihood. Or, if you take "minimization" to be a synonym of "numerical minimization", then the pull method that is actually being introduced by this PR also doesn't fall into that category.

On your second comment: I'll make those copies then. The good thing is that the potentially large history entry (`fit_history`, in case the scripts are also using that one) remains unchanged.

bourdeet:
I don't know if anyone else is still checking the outcome of this PR on their analysis, but for my part, I've ran both asimov and sensitivity test and they both yielded identical results:

- Asimov, using the master branch: https://icecube.wisc.edu/~bourdeet/thomas_PR_checks/test_thomas_pr_asimov_master.pdf
- Asimov, this PR: https://icecube.wisc.edu/~bourdeet/thomas_PR_checks/test_thomas_pr_asimov.pdf

- sensitivity, master branch: https://icecube.wisc.edu/~bourdeet/thomas_PR_checks/test_thomas_pr_sensitivity.pdf
-sensitivity, this PR: https://icecube.wisc.edu/~bourdeet/thomas_PR_checks/test_thomas_pr_sensitivity_master.pdf

So for my part, I'm satisfied with this PR, especially if the `minimizer_` key is kept as a copy for backward compatibility :)

thehrh:
Thanks for the comparisons. I'm looking at what is going on with the number of minimiser iterations reported, given the consistent offset of one. Then I'll report back so that we can hopefully merge this branch.

philippeller:
@thehrh any news on the last remaining issue? Please let @jllanfranchi and/or me know when this is ready to be merged!

philippeller:
Don't ask me why, but it added my pisa4 dev changes once i merged this into my pisa4 dev branch....

philippeller:
Anyhow, I think we can declare this PR as stale, and I took over the changes into pis4 anyhow, so gonna close this

PR #588, opened 2020-03-19
Hypersurface test cleanup

atrettin:
When I started fixing the unit test I ended up doing a huge cleanup operation on the hypersurface code. I apologize for the huge changeset, a lot comes just from running autopep8 and finally fixing all the weird spacing, and from re-flowing comment/doc-string paragraphs that were too wide.

Things that actually change functionality are:
- Fix unit tests: They now actually test something and conform to naming and logging conventions. Passes now in `run_unit_tests.py`.
- Remove smoothing feature: I believe that it will do more harm than good being in there. If you want smoothing, put it into your pipeline and _then_ fit hypersurfaces on it. It has no business being part of the HS fit where it will just produce wrong numbers. 
- Remove parts that were meant for Numba expansion in the future, because we won't do that.


atrettin:
Fixes the hypersurface part of issue 586.

atrettin:
Since the numbers tested are the results of a minimization, they are quite sensitive to the input precision. I found the difference between true and recovered coefficients to be O(1e-5) in fp32 and O(1e-12) in fp64. To me that would be acceptable, but I can also try to fiddle with the iminuit settings to make it more reliable.

jllanfranchi:
I wouldn't try to modify things further, we just have to accept that things behave differently if we run fast (fp32) vs slower/more precise (fp64).

And note that those orders-of-magnitude should be what we defined for `ALLCLOSE_KW`. If using that doesn't give you agreement, let me know or just modify accordingly to get the test to pass (that many digits is approx. 4 bytes and 8 bytes of precision; more advanced comparisons can be done not in decimal but taking the binary representation into account. But we'll only go there if absolutely necessary, or else just hit it with a big hammer and allow for one less digit in the comparison compared to what we've already defined.)

atrettin:
Ah ok, then the differences found are _just_ above the tolerance. In fp32, it fails the test with differences of `[[7.915497e-05]
 [7.915497e-05]
 [7.915497e-05]]`
And in fp64, when I use ALLCLOSE_KW, it also fails ( in a different place of the test) with differences of
`[[-1.61348712e-12  1.53743684e-12]]`, which ist just a smidge above 1e-12. 
What I could do it take rtol from ALLCLOSE_KW and multiply it by 10, maybe?

jllanfranchi:
Yeah, I think that's fair. That's the "big hammer" approach, but I think even with one less digit (factor of 10) for comparison, it should still be good enough to verify that it's working as you expect.

atrettin:
Should be fixed now...

PR #590, opened 2020-03-20
Container tests

philippeller:


pisa/core/container.py
```diff
@@ -545,24 +546,32 @@ def test_container():
...
     binning_y = OneDimBinning(name='y', num_bins=10, is_lin=True, domain=[0, 100])
     binning = MultiDimBinning([binning_x, binning_y])
-    #print(binning.names)
-    print(container.get_binned_data('x', binning).get('host'))
-    print(Container.unroll_binning('x', binning).get('host'))
 
-    # array
-    print('original array')
-    print(container.get_array_data('w').get('host'))
+    logging.trace('Testing container amd translation methods')
```
jllanfranchi: sp

pisa/core/container.py
```diff
@@ -545,24 +546,32 @@ def test_container():
...
+
+    bx = container.get_binned_data('x', binning).get('host')
+    m = np.meshgrid(binning.midpoints[0].m,  binning.midpoints[1].m)[1].ravel()
+    assert np.allclose(bx, m)
+
+    # array repr
+    array_weights = container.get_array_data('w').get('host')
+    assert np.allclose(array_weights, w)
```
jllanfranchi: use `**ALLCLOSE_KW` (`from pisa.utils.comparisons import ALLCLOSE_KW`) here to get comparisons that adjust precision to `FTYPE` (here and below); also, display the contents of each item (test, ref) if there's a failure so the user doesn't have to hack any code to start narrowing down the issue, as that will be present in the output logs of the unit test runner. E.g. my go-to since f-strings are so succinct:

```python
assert np.allclose(array_weights, w), f'test:\n{array_weights}\n!= ref:\n{w}'
```
(or similar)

philippeller:
Ok, implemented requested changes. Gonna close this PR and reopen against your branch

PR #593, opened 2020-03-22
Cleanup numba_osc code

jllanfranchi:
* add unit test for untested function
* add wrappers so all tests can be run whether TARGET={cpu,cuda}
* define reference test values to test output against (some based on current outputs). while not perfect (maybe we have a bug now?) at least we can test that code doesn't change from now
* add docstrings
* black code formatting
* NOTE: odd issue where it kinda looks like i4 has to be used with FTYPE=float32, and i8 with FTYPE=float64? I didn't hunt this issue down much further, but it works when I use that so I'm sticking with it.
* NOTE: compilation time is rather slow! Possibly due to the wrapper functions. If this is an issue, we can make the wrapper functions closures (define within the actual test functions) such that importing the module should still be sufficiently fast for actual real-world usage (only testing will be slow)


thehrh:
How do we coordinate the merging of this and PR #567 with @ellohfin? The changes to the core of the oscillation-probability calculation in #567 are rather moderate -> do we try and merge those first? I would be willing to help a bit with the cleaning up once that's done.

jllanfranchi:
I basically only changed formatting the actual kernel functions, in addition to adding more / making consistent the docstrings. I had a quick look at the other code, and it looks like it could be merged without too much difficulty. I think a good thing to start with before we do this, since I now have unit testing framework for this code, is to define some useful test cases with nsi_eps != 0 in the code I have here. We save the results, then merge the code from @ellohfin into this code, then run the tests again and hopefully the results are consistent (assuming those changes don't fundamentally change the calculation).

ellohfin:
I'd assume that merging #567 first might be easiest, since there is not as much changing there? Also, apart from one test still failing, I'd assume to be basically done with that.

philippeller:
I found that explicitly calculating all 3 flavour vacuum osc probs seems to work fine:
i.e. setting the loop in L.93 in `pisa/stages/osc/prob3numba/numba_osc_kernels.py` to run over all 3 flavours, and commenting out the `1.0 - osc_probs_local[ista, 0] - osc_probs_local[ista, 1]` produces sensible probabilities.

Of course now the test fails, as it has wrong reference values:
```<< PISA is running in double precision (FP64) mode; numba is running on GPU >>                                                                                                              [   ERROR] propagate_scalar_vacuum: key: 'probability' :: test:                                                                                                                             [[ 0.98517492169974585 -0.01482507830025410 -0.01482507830025410]                                                                                                                            [-0.48420178468166286  0.51579821531833714 -0.48420178468166286]                                                                                                                            [-0.47951617325806195 -0.47951617325806195  0.52048382674193805]]                                                                                                                          != ref:                                                                                                                                                                                     [[ 0.98517492169974585 -0.01482507830025410  0.02965015660050825]                                                                                                                            [-0.48420178468166286  0.51579821531833714  0.96840356936332572]                                                                                                                            [-0.47951617325806206 -0.47951617325806206  1.95903234651612390]]  ```

But what is output looks reasonable.

So the culprit is the `prob_tau = 1 - prob_e - pron_mu` part.
Still it makes absolutely no sense that setting `c = 1 - a -b` ends up with `a + b + c != 1` (_very_ not-equal to 1 that is). Sounds like a bug somewhere deeper. We could try to come up with a minimal example....

philippeller:
Or maybe not.....doesn't work again :(

jllanfranchi:
@ellohfin @thehrh @philippeller when you get a chance, please check the physics correctness of this branch. It's passing Github CI, and is within 1.5% accuracy running on GPU fp32 to passing oscillations tests there. We can worry about GPU stuff later.

Thanks!

philippeller:
I checked out your branch and run the Asimov significance for the dragon nutau analysis - results is unchanged. So good from my side to merge!

jllanfranchi:
@philippeller thanks for testing!

@thehrh @ellohfin I'm going to go ahead and merge, but if you could, please look over the merged code within `numba_osc_kernels.py` when you have a chance to make sure I got the NSI stuff correct, which we haven't test.

Also, in a subsequent PR, I would appreciate if you could come up with one or more sets of values for `mat_pot` that make sense for unit tests so we can make sure that results are at least consistent over time for NSI. (See, e.g., test cases defined here: https://github.com/IceCubeOpenSource/pisa/pull/593/files#diff-cb03d036d25a605f35d7b1a665b6d383R107-R128

(I'll add an issue for this to remind us after I merge/close this PR.)

PR #597, opened 2020-03-24
Scaled poisson fluctuation

atrettin:
The current implementation of the `poisson` fluctuation method cannot be used to fluctuate MC events in a meaningful way. Furthermore, the fluctuation does not have the standard deviation that is stored in the Map, and instead fudges the standard deviation in a way the user might not expect. 

This PR implements the Scaled Poisson Process derived in [Bohm & Zech 2018](https://arxiv.org/abs/1309.1287) and applies it when the stored standard deviations in a map are not sqrt(N). This process ensures that the mean and standard deviation are correctly observed and is a reasonable approximation from the fluctuation that can be expected from a Compound Poisson Process that produces our MC expectations. 

If a map has errors of sqrt(N), as is the case after `set_poisson_errors()` has been called, there is no difference between the new and old behavior.

A warning is issued when a map has no associated errors, as the behavior is technically undefined. In that case, we fall back to the old behavior, but warn the user about what is happening.

### DIFF
--- a/pisa/core/map.py
+++ b/pisa/core/map.py
@@ -31,7 +31,7 @@ from uncertainties import unumpy as unp
 
 from pisa import ureg, HASH_SIGFIGS
 from pisa.core.binning import OneDimBinning, MultiDimBinning
-from pisa.utils.comparisons import normQuant, recursiveEquality
+from pisa.utils.comparisons import normQuant, recursiveEquality, ALLCLOSE_KW
 from pisa.utils.flavInt import NuFlavIntGroup
 from pisa.utils.hash import hash_obj
 from pisa.utils import jsons
@@ -961,20 +961,29 @@ class Map(object):
             random_state = get_random_state(random_state, jumpahead=jumpahead)
             with np.errstate(invalid='ignore'):
                 orig_hist = self.nominal_values
+                sigma = self.std_devs
+
                 nan_at = np.isnan(orig_hist)
+                zero_at = orig_hist == 0.
                 valid_mask = ~nan_at
+                variance_valid = sigma[valid_mask]**2
+
+                if np.allclose(variance_valid, orig_hist[valid_mask], **ALLCLOSE_KW):
+                    scale_factor = 1.
+                else:
+                    scale_factor = variance_valid/orig_hist[valid_mask]
+                poisson_lambda = orig_hist[valid_mask]/scale_factor
 
                 hist_vals = np.empty_like(orig_hist, dtype=np.float64)
                 hist_vals[valid_mask] = poisson.rvs(
-                    orig_hist[valid_mask],
+                    poisson_lambda,
                     random_state=random_state
                 )
+                hist_vals[valid_mask] *= scale_factor
                 hist_vals[nan_at] = np.nan
-
-                error_vals = np.empty_like(orig_hist, dtype=np.float64)
-                error_vals[valid_mask] = np.sqrt(orig_hist[valid_mask])
-                error_vals[nan_at] = np.nan
-            return {'hist': unp.uarray(hist_vals, error_vals)}
+                hist_vals[zero_at] = 0.
+                # this implementation ensures that the standard deviation is unchanged
+            return {'hist': unp.uarray(hist_vals, sigma)}
 
         elif method == 'gauss+poisson':
             random_state = get_random_state(random_state, jumpahead=jumpahead)
--- a/pisa/core/map.py
+++ b/pisa/core/map.py
@@ -943,9 +943,9 @@ class Map(object):
             'gauss+poisson'. Strings are case-insensitive and whitespace is
             removed.
             The 'scaled_poisson' method implements a Scaled Poisson Process, which is 
-            a better approximation to the true distribution when the bin counts are 
-            the result of a Poisson process with weighted events[1] (as is the case in MC
-            maps). The fluctuated maps are guaranteed to have the same mean and standard
+            a better approximation than a normal distribution to the true distribution 
+            of bin counts that are the result of a Poisson process with weighted events[1].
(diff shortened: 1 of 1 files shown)


thehrh:
I don't think I understand what we're trying to achieve here. Is the idea that applying Poisson fluctuations to a map---without having called `set_poisson_errors()` before---will use the MC uncertainty instead? This seems like it will break existing code, or am I overlooking something? Are these modifications supposed to replace the `gauss` method we have implemented, which applies normal variations to the bin counts (with std. devs. composed of the various MC events' weights)?

thehrh:
If any of my suspicions apply (in some way or another), I think we should instead come up with a new "fluctuation" method for this approach. The "poisson" approach should stay naive and not know about how the template was generated, imho.

atrettin:
> I don't think I understand what we're trying to achieve here. Is the idea that applying Poisson fluctuations to a map---without having called `set_poisson_errors()` before---will use the MC uncertainty instead? This seems like it will break existing code, or am I overlooking something? Are these modifications supposed to replace the `gauss` method we have implemented, which applies normal variations to the bin counts (with std. devs. composed of the various MC events' weights)?

These modifications do not replace the 'gauss' method.

The goal is to fluctuate maps in a way that approximates a Compound Poisson Process, which is the process of adding weighted events. This process conserves the error that is stored in the Map, i.e. the fluctuated values have the same mean and standard deviation that is originally put into the Map. The current implementation overrides those errors. 

It is possible that code could break if it relied on the fact that the error is overridden. I can change it to be a separate fluctuation method `scaled_poisson` if you would be happier with that.

thehrh:
Thanks for the clarifications. Still, to me what you're trying to achieve sounds a lot like our `gauss` fluctuation method (But perhaps in a more correct way, since you're not assuming the Gaussian limit of adding weighted events? Please correct me if I'm way off).

I personally dislike the idea of the Poisson process in future by default using the uncertainties stored with the maps, and not the actual, sqrt(n), Poisson uncertainty of the template. Is the latter what you're referring to as "fudging of numbers in the background without telling the user about it"?

thehrh:
I did  take a look at your code but I haven't looked up the method you're referring to, so apologies if I'm misunderstanding.

atrettin:
The reason for this endeavor is to test if our Gaussian assumption that goes into the modified chi-square loss function holds. While the normal distribution has the correct mean and standard deviation, it is otherwise a bad approximation to a Compound Poisson Process. Bohm&Zech show that the Scaled Poisson Process is a much better approximation. While I did not delete the 'gauss' method of fluctuation, the scaled poisson method should be used instead to simulate Monte Carlo uncertainty.

Yes, it seems like we have a very different perspective on the same thing. To me, it seems correct that the maps should be fluctuated in a way that respects the standard deviations given as errors, and it seems like "fudging the numbers" when the output of the `.fluctuate` method has different errors than the one that was given to it.

You seem to come from the perspective that the expectation of the user is that Poisson fluctuations should always have errors of sqrt(N), regardless of the standard deviations stored in the Map.

The answer to the question on what is correct here is a question of convention and what a user's expectation of the behavior is... again, if you would be more comfortable with that, I can make this a separate method with the name 'scaled_poisson' and then we can be absolutely  sure that there is no interference with existing code.

thehrh:
Ok, cool, then we're on the same page. My preference is a new method. Also because that's trivially backward compatible.

atrettin:
I have addressed the compatibility and documentation issues. The method 'scaled_poisson' is now separate, and 'poisson' is left unchanged.

pisa/core/map.py
```diff
@@ -975,6 +984,43 @@ def fluctuate(self, method, random_state=None, jumpahead=0):
...
+                    sigma[valid_mask] = np.sqrt(orig_hist[valid_mask])
+                variance_valid = sigma[valid_mask]**2
+
+                if np.allclose(variance_valid, orig_hist[valid_mask], **ALLCLOSE_KW):
+                    scale_factor = 1.
+                else:
+                    scale_factor = variance_valid/orig_hist[valid_mask]
+                poisson_lambda = orig_hist[valid_mask]/scale_factor
+
+                hist_vals = np.empty_like(orig_hist, dtype=np.float64)
```
thehrh: not `dtype=FTYPE`?

jllanfranchi: good point ... or even just leave off the dtype, as `empty_like` will pick up the dtype from `orig_hist`

atrettin: Then I'll change it in the original 'poisson' implementation as well, it does the same.

atrettin: Unless there was a reason to do it like that in the original?

jllanfranchi: nope, I don't think so

pisa/core/map.py
```diff
@@ -939,9 +939,14 @@ def fluctuate(self, method, random_state=None, jumpahead=0):
         Parameters
         ----------
         method : None or string
-            Valid strings are '', 'none', 'poisson', 'gauss', or
+            Valid strings are '', 'none', 'poisson', 'scaled_poisson', 'gauss', or
             'gauss+poisson'. Strings are case-insensitive and whitespace is
             removed.
+            The 'scaled_poisson' method implements a Scaled Poisson Process, which is 
+            a better approximation to the true distribution when the bin counts are
```
thehrh: Better approximation compared to what? Which "true" distribution? You mean the true distribution of the random variates due to a finite number of MC events, right? I keep coming back to this because pisa originally wasn't intended to be able to reweight MC events. This shouldn't hold up the PR, but we should work on documentation (also, and in particular, for the other methods) at some point.

pisa/core/map.py
thehrh: yep, probably an oversight

thehrh:
looks good, thanks!

PR #600, opened 2020-03-27
Implementation of DIS uncertainties

philippeller:
This is from the work of Maria / Juan Pablo
Implemented a stage applying the corrections to the events with a new paramter `dis_csms`


pisa/stages/xsec/dis_sys.py
```diff
@@ -0,0 +1,213 @@
...
+            #
+            # Calculate variation of total cross section
+            #
+
+            lgE_min = 2. if self.extrapolation_type == 'constant' else 1.68
+            mask = lgE <= lgE_min
+
+            w_tot = np.ones_like(lgE)
+            
+            poly_coef = extrap_dict[nu][current]['poly_coef']
```
jllanfranchi: I think this block of code could be made more clear in what it's doing, more concise, and faster at the same time:

```python
poly_coef = extrap_dict[nu][current]['poly_coef']
lin_coef = extrap_dict[nu][current]['lin_coef']

if self.extrapolation_type == 'higher':
    w_tot = np.polyval(poly_coef, lgE)
else:
    poly_valid_mask = lgE >= lgE_min
    poly_invalid_mask = ~poly_valid_mask

    w_tot[poly_valid_mask] = np.polyval(poly_coef, lgE[poly_valid_mask])

    if self.extrapolation_type == 'constant':
        w_tot[poly_invalid_mask] = np.polyval(poly_coef, lgE_min)  # note Numpy broadcasts
    elif self.extrapolation_type == 'linear':
        w_tot[poly_invalid_mask] = np.polyval(lin_coef, lgE[poly_invalid_mask)
    else:
        raise ValueError('Unknown extrapolation type "%s"'%self.extrapolation_type)
```

pisa/stages/xsec/dis_sys.py
```diff
@@ -0,0 +1,213 @@
...
+
+    @profile
+    def apply_function(self):
+        dis_csms = self.params.dis_csms.m_as('dimensionless')
+
+        for container in self.data:
+            apply_dis_sys(
+                container['dis_correction_total'].get(WHERE),
+                container['dis_correction_diff'].get(WHERE),
+                dis_csms,
```
jllanfranchi: `FTYPE(dis_csms)`

pisa/stages/xsec/dis_sys.py
```diff
@@ -0,0 +1,213 @@
...
+                out=container['weights'].get(WHERE),
+            )
+            container['weights'].mark_changed(WHERE)
+
+
+if FTYPE == np.float64:
```
jllanfranchi: Avoiding copy-paste errors got easier since f-strings; I'd like to propose that people use the following convention for these signatures from now on:
```python
FX = 'f8' if FTYPE == np.float64 else 'f4'

@guvectorize([f'({FX}, {FX}, {FX}, {FX}[:])'], '(),(),()->()', target=TARGET)
 def apply_dis_sys(...)
```

pisa/stages/xsec/dis_sys.py
```diff
@@ -0,0 +1,213 @@
...
+            elif current == 'NC' and container['nubar'] > 0:
+                weight_func = wf_nunc
+            elif current == 'NC' and container['nubar'] < 0:
+                weight_func = wf_nubarnc
+
+            w_diff[~mask] = weight_func.ev(lgE[~mask], bjorken_y[~mask])
+            w_diff[mask] = weight_func.ev(w_diff[mask] * lgE_min, bjorken_y[mask])
+
+            # mke centered arround 0, and set to 0 for all non-DIS events
+            w_diff -= 1
```
jllanfranchi: combine into one operation (L182+L183)

pisa/stages/xsec/dis_sys.py
```diff
@@ -0,0 +1,213 @@
...
+            container["dis_correction_total"] = w_tot
+
+            #
+            # Calculate variation of differential cross section
+            #
+
+            lgE_min = 2.
+            mask = lgE <= lgE_min
+
+            w_diff = np.ones_like(lgE)
```
jllanfranchi: empty_like

pisa/stages/xsec/dis_sys.py
```diff
@@ -0,0 +1,213 @@
...
+
+        self.extrapolation_type = extrapolation_type
+
+    @profile
+    def setup_function(self):
+
+        extrap_dict = from_file('cross_sections/tot_xsec_corr_Q2min1_isoscalar.pckl')
+
+        # load splines
+        wf_nucc =    from_file('cross_sections/dis_csms_splines_flat/NuMu_CC_flat.pckl')
```
jllanfranchi: no value to creative spacing

pisa/stages/xsec/dis_sys.py
```diff
@@ -0,0 +1,213 @@
...
+            w_tot *= dis
+          
+            container["dis_correction_total"] = w_tot
+
+            #
+            # Calculate variation of differential cross section
+            #
+
+            lgE_min = 2.
+            mask = lgE <= lgE_min
```
jllanfranchi: again, use a meaningful mask name, and also again, not clear why the endpoint isn't included in valid range? Also, need both masks multiple times, so avoid multiple negations of the entire array by defining both:
E.g.,
```python
spline_valid_mask = lgE >= lgE_min
spline_invalid_mask = ~spline_valid_mask
```

pisa/stages/xsec/dis_sys.py
```diff
@@ -0,0 +1,213 @@
...
+            w_tot -= 1
+            w_tot *= dis
+          
+            container["dis_correction_total"] = w_tot
+
+            #
+            # Calculate variation of differential cross section
+            #
+
+            lgE_min = 2.
```
jllanfranchi: retrieve from the file, don't hard-code here

pisa/stages/xsec/dis_sys.py
```diff
@@ -0,0 +1,213 @@
+"""
+Stage to apply pre-calculated DIS uncertainties
```
jllanfranchi: please add a reference so others know where all of this comes from (a presentation, a paper, etc.); also references to author(s) are nice here so people know who to ask. And not just you, but presumably Maria and Juan Pablo. You can answer coding questions, but they would be the first people to turn to to answer physics questions. Readers of the code need to know those things (and the text you put on the PR will be difficult if impossible to find).

philippeller:
This plot is showing the effect on the oscnext sample as a function of true energy and inelastcicity, directly from this stage when comparing nominal (`dis_csms = 0`) vs a variation (`dis_csms = 1`)

image

pisa/stages/xsec/dis_sys.py
```diff
@@ -0,0 +1,213 @@
...
+
+            lgE = np.log10(container['true_energy'].get('host'))
+            bjorken_y = container['bjorken_y'].get('host')
+            dis = container['dis'].get('host')
+
+            #
+            # Calculate variation of total cross section
+            #
+
+            lgE_min = 2. if self.extrapolation_type == 'constant' else 1.68
```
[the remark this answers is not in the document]
philippeller: Unfortunately I can only find the number 1.68 in the file, but not 2......those must be some hand chosen, magic values...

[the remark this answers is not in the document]
jllanfranchi: Right, I see 1.68 is in `tot_xsec_corr_Q2min1_isoscalar.pckl`, so that should be dynamically loaded. Can you find out from @jpyanez and/or Maria why the 2 magic number is chosen for constant extrapolation?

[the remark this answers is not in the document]
philippeller: Talking to juan pablo, this really is just a choice, and they tried different values between 50 and 100 GeV. So I now made this an argument to the function that defaults to 100

pisa/stages/xsec/dis_sys.py
philippeller: done

pisa/stages/xsec/dis_sys.py
philippeller: the ones are needed later on

pisa/stages/xsec/dis_sys.py
philippeller: done

philippeller:
All comments addressed

pisa/stages/xsec/dis_sys.py
jllanfranchi: Ah I see, you're multiplying N 1's by `lgE_min` to obtain N copies of `lgE_min`. Broadcasting applies just the same here as in numpy, so all you have to do is pass `lgE_min` as the first arg; below can become simply:
```python
w_diff = np.empty_like(lgE)
w_diff[valid_mask] = weight_func.ev(lgE[valid_mask], bjorken_y[valid_mask])
w_diff[extrapolation_mask] = weight_func.ev(lgE_min, bjorken_y[extrapolation_mask])
```
This avoids creation of an extra array and a bunch of multiplies, but since it's in the setup function that might not matter. Change if you want, then feel free to merge.

philippeller:
Oh hey, you;re right! Didn't expect that to work but it does. So i changed it, it's simpler. Gonna push that change now and then merge - thanks for the review!

PR #601, opened 2020-03-31
add tentative nsi mat_pot test case

### LINKED ISSUE #599: Add NSI unit tests to prob3numba/numba_osc_tests.py
Could use one or more sets of values for `mat_pot` that make sense for unit tests so we can make sure that results are at least consistent over time for NSI. (See, e.g., test cases defined here: https://github.com/IceCubeOpenSource/pisa/blob/48bdc371cc230ef4e8a3afa29c1ad58aac606a6c/pisa/stages/osc/prob3numba/numba_osc_tests.py#L107-L128)

That file can be called directly as a script from the command line with the flag `--define-as-ref`, and should be done so running on CPU (we are making CPU the reference) both with PISA_FTYPE=fp32 and PISA_FTYPE=fp64. This should generate the test data files that we can use into the future for whatever test case(s) you come up with, and we can selectively check in just those files you generate for those tests.

ellohfin:
Fixes #599


jllanfranchi:
First, note that development is easier if you start a task by git checkout i3os/master, then create a branch from that: git checkout -b feature_name. After we merge then delete that feature branch.

Second, produce reference files also in fp32 (you'll notice this is why unit tests are failing)

ellohfin:
Hope creating this PR made my questions on issue #599 clearer

jllanfranchi:
Yes, that does make sense! Thanks for adding these additional unit tests. Squash-merging now (so note your `master` branch will be even _more_ out of sync with `icecubeopensource/master` than before, and next PR will look even crazier with number of commits...)

PR #603, opened 2020-04-01
fix the distribution_maker get_outputs function to properly handle more than 2 pipelines

bourdeet:
This is a simple fix on the `get_outputs` function of distribution maker, in the case where we want to return the sum of all maps. 

previous behavior (using the python `reduce`) didn't work once you added a third pipeline, because the sum of two MapSet is a Map, and not a MapSet.

### DIFF
--- a/pisa/core/distribution_maker.py
+++ b/pisa/core/distribution_maker.py
@@ -203,7 +203,8 @@ class DistributionMaker(object):
         outputs = [pipeline.get_outputs(**kwargs) for pipeline in self] # pylint: disable=redefined-outer-name
         if return_sum:
             if len(outputs) > 1:
-                outputs = reduce(lambda x, y: sum(x) + sum(y), outputs)
+                outputs = sum([sum(x) for x in outputs])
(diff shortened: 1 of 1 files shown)


philippeller:
Great!
Only thing: remove the now unused import of `reduce`

bourdeet:
Here you go!

pisa/core/distribution_maker.py
```diff
@@ -203,7 +202,8 @@ def get_outputs(self, return_sum=False, sum_map_name='total',
         outputs = [pipeline.get_outputs(**kwargs) for pipeline in self] # pylint: disable=redefined-outer-name
         if return_sum:
             if len(outputs) > 1:
-                outputs = reduce(lambda x, y: sum(x) + sum(y), outputs)
+                outputs = sum([sum(x) for x in outputs])
```
jllanfranchi, line 205: No need to create a list and then run `sum` over the list;, `sum` works directly with an iteration:
```python
sum(sum(x) for x in outputs)
```

Issue #596, opened 2020-03-24
Modules fail to import when using gpu (CUDA) mode


jllanfranchi:
- [ ] `pisa.stages.flux.pi_mceq_barr`; authors listed: @ts4051 @philippeller @blotsd 
- [x] `pisa.stages.absorption.pi_earth_absorption` no authors; initial commit @atrettin
  - Fixed with with PR #605 / https://github.com/IceCubeOpenSource/pisa/commit/d6d787c3946a5be6e6226c40dd43d149d717ae24

philippeller:
as for pisa4, we will not support cuds in this stage anymore

PR #605, opened 2020-04-03
Make earth absorption stage compatible with GPU target

atrettin:
This PR addresses issue #596 . The unsupported functions `np.dot` and `np.exp` are replaced so that both 'cpu' and 'gpu' targets work with this stage. I checked that both import without errors and give the same result.

### DIFF
--- a/pisa/stages/absorption/pi_earth_absorption.py
+++ b/pisa/stages/absorption/pi_earth_absorption.py
@@ -202,6 +202,7 @@ class pi_earth_absorption(PiStage):
                                                              container['nubar'],
                                                              container['true_energy'].get(WHERE)
                                                             )
+            container['xsection'].mark_changed(WHERE)
             calculate_survivalprob(container['rho_int'].get(WHERE),
                                    container['xsection'].get(WHERE),
                                    out=container['survival_prob'].get(WHERE)
--- a/pisa/stages/absorption/pi_earth_absorption.py
+++ b/pisa/stages/absorption/pi_earth_absorption.py
@@ -184,6 +184,7 @@ class pi_earth_absorption(PiStage):
                                      container['densities'].get(WHERE),
                                      out=container['rho_int'].get(WHERE)
                                     )
+            container['rho_int'].mark_changed(WHERE)
         # don't forget to un-link everything again
         self.data.unlink_containers()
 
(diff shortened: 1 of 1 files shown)


jllanfranchi:
Quick glance on my phone this looks fine, but please keep the comment identifying the 1e5 constant, we don't like magic numbers just appearing in the code that subsequent users may be unclear on

atrettin:
sure, I added the comment back in

pisa/stages/absorption/pi_earth_absorption.py
```diff
@@ -133,7 +149,7 @@ def setup_function(self):
                                              'nuebar_nc', 'numubar_nc', 'nutaubar_nc'])
 
         for container in self.data:
-            self.layers.calcLayers(container['true_coszen'].get(WHERE))
+            self.layers.calcLayers(container['true_coszen'].get('host'))
```
jllanfranchi, line 152: Is this necessary to run on gpu, to force this to be done on cpu? And is this a performance issue? I don't fully understand it but obviously this stands out

atrettin: The layers module fails on the GPU target apparently, and this is done the same way in the prob3 stage (but only once during setup).

jllanfranchi: Fair enough. Don't you have to mark changed after an operation on the smart array, too?

atrettin: The layers module seems to be a host-only function. Since the containers for densities and distances are only created once, they don't have to be marked changed. However, it was still a good call, because the `rho_int` container does indeed change and needs to be marked as such.

jllanfranchi:
@philippeller anything to add? If not, feel free to merge...

pisa/stages/absorption/pi_earth_absorption.py
philippeller: I'd still advocate to marke stuff as changed no matter if it is a special case like this where we can get away without doing it. It is just good practice and it does not create any overhead. The copy to and from device are lazy evaluations, so they only get executed when the array is needed in a subsequent calculation.

atrettin:
I marked the containers changed on creation as well.

PR #606, opened 2020-04-06
Use flag to determine whether baseline flux comes from MCEq or pi_hon…

blotsd:
…da_ip stage

### DIFF
--- a/pisa/stages/flux/pi_mceq_barr.py
+++ b/pisa/stages/flux/pi_mceq_barr.py
@@ -32,6 +32,7 @@ class pi_mceq_barr(PiStage):
     Parameters
     ----------
     table_file : pickle file containing pre-generated tables from MCEq
+    use_MCEq_nom_flux : True or False, use MCEq for nominal flux calculation
 
     params : ParamSet
         Must exclusively have parameters: .. ::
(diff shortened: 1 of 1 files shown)


philippeller:
Hey, wouldn't it make more sense to have the nominal flux calculation in a separate stage? So it could be transparently switched out between honda and MCEq and no need for flags or anything like it

blotsd:
Yes, I think that would be a cleaner solution. My only concern is the loading of the spline tables - this would be done twice then, once in the nominal flux calculation stage and once again in the flux systematic stage... unless you can guide me or provide an example of how we could load once and access across multiple stages

blotsd:
I think there is also a need to re-write the honda_ip and barr_simple stages - right now honda_ip calculates both nu and nubar flux for all species (regardless of whether they are nu or nubar) and so that's what barr_simple works from as well. I don't see why this is necessary? And it makes compatibility with mceq_barr stage annoying. 

How would you feel about just keeping the flag I implemented for now, and then in a few weeks I'll totally revamp the flux stages to make them all cleaner and easier to  mix and match? This would include a new, separate stage for MCEq-based nominal flux calculation.  I am worried about meeting oscnext deadlines otherwise....

philippeller:
fine with me. We don't need the _best_ possible solution right now, can always improve. Just need to make sure the current one does what it's supposed to

jllanfranchi:
Any reason to not merge this, @philippeller ?

philippeller:
Yes, *don't* merge - @sblot wanted to close this PR...there is a new one coming

blotsd:
Yup, sorry closing now.

PR #607, opened 2020-04-14
New data loading functionality

ts4051:
Two new data loading features:

1) Specify a list of input files rather than a single one, which are loaded as if they were a single large file. Example use case if to load detector data which is stored as one season per file (makes it easy to test individual seasons).

2) Only load specified variables from the HDF5 files. Allows HDF5 files containing supplementary data like control variables to be made, but those variables are not loaded when not required (avoids high memory costs).


jllanfranchi:
Note that unit checks are failing (red X...); please fix your code before any review / merging will be done. Also, note that we enforce strict style guidelines for all code that goes into PISA's core, so be prepared for a little more nit-picking from me on anything you do there.

pisa/core/events_pi.py
```diff
@@ -66,6 +66,25 @@
...
+    Helper function for appending multiple dicts of arrays (e.g. from 
+    multiple input files) into a single dict of arrays 
+    '''
+    if isinstance(val, Mapping):
+        for key2, val2 in val.items() :
+            if key not in sdict :
+                sdict[key] = OrderedDict()
+            append_arrays_dict(key2, val2, sdict[key])
+    else :
+        assert isinstance(val, np.ndarray), "'%s' is not an array, is a %s" % (k, type(v))
```
ellohfin, line 81: seems like this crashes a test because you use "k" and "v" instead of "key" and "value"

ts4051:
OK tests are passing now.

ts4051:
I am closing this PR as the changes have already been merged to master as part of PR #626.

Issue #608, opened 2020-04-14
Stage configuration arguments are parsed as Quantity


atrettin:
When a configuration file of a pipeline is parsed, stage initialization arguments not meant for optimization are parsed as quantities with units, even when only a simple number is given. When a config file contains ```seed = 42```, it currently gets ```Quantity(42.0, dimensionless)``` as an argument into its ```__init__``` function, when it should get the number 42 as an integer.

jllanfranchi:
Fixed with PR #609 / commit 8b6d88627e46519b1431effe4626eb0ac0085678

PR #609, opened 2020-04-14
Attempt to fix issue #608,  parsing non-param args in sensible way

jllanfranchi:
Attempt to fix issue #608 :

If `[^a-z]units\.[a-z]+` regex is found in a value being parsed, try to parse as a Pint quantity (possibly with errors via uncertainties). If that fails, or if that string wasn't found, parse as a literal value (i.e., NOT as a Pint quantity).


atrettin:
Tested the PR on a stage I was developing. All seems to work as intended. :)

PR #610, opened 2020-04-14
add MC fluctuation stage

atrettin:
This is a simple stage to simulate the impact of MC uncertainties. The basic idea is that every event is interpreted as a one-of-a-kind observation. During fluctuation, every event is re-weighted by a Poisson factor with an expectation value of one. The resulting distribution is the closest approximation to the true distribution of MC weights, which is unknown. This is also known as the "Poisson Bootstrap Distribution" (PBD).

The technical advantage of this implementation is that the rest of the PISA pipeline can run unaffected, making it simple to integrate into ensemble tests.

### DIFF
--- a/pisa/stages/utils/pi_fluctuation.py
+++ b/pisa/stages/utils/pi_fluctuation.py
@@ -1,0 +1,100 @@
+"""
+Stage to fluctuate MC events in a pipeline.
+
+Each event is interpreted as a unique observation and its weight multiplied by a Poisson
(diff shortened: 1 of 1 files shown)


atrettin:
After more discussion with Thorsten Glüsenkamp, who did extensive work on the topic of MC uncertainties, I have come to the conclusion that this topic is difficult and it would take a significant amount to show first that this is the right thing to do. I'm closing the PR for now and put the topic on the back burner for things to consider in a later analysis.

PR #611, opened 2020-04-16
add kfold splitting stage

atrettin:
This stage splits the data into k-folds using the `KFold` class from `scikit-learn`. It then selects the "test" indices from one of the folds and sets the weights of all other events to zero. This allows one to run a pipeline with one part of an MC set and fit it back with another. Optionally, weights can be multiplied by the number of splits to normalize the final rates.

### DIFF
--- a/pisa/stages/utils/pi_kfold.py
+++ b/pisa/stages/utils/pi_kfold.py
@@ -1,0 +1,128 @@
+"""
+Make K-folds of data.
+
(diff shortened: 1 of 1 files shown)


philippeller:
It looks ok on first inspection! I'm proposing to merge that.....any onjections?

atrettin:
There is a "mark_changed" that I didn't push yet, one sec...

PR #612, opened 2020-04-21
remove uniform binning checks

atrettin:
Removes the check that requires bins to be spaced uniformly. Bins can now be merged and maps still be plotted with the correct axis scaling (log vs. lin, bin midpoints).


jllanfranchi:
PR #615 should subsume this (I started with your commit here and modified from there)

PR #613, opened 2020-04-21
new MCEq stage with reduced functionality

blotsd:
NO baseline flux calculation and reduced number of Barr parameters expected


blotsd:
Accidentally made PR with both previous commits. Cancelling this one and trying again

PR #615, opened 2020-04-21
`OneDimBinning`: `is_lin`, `is_log` describe binning behavior, NOT regularity in either space

jllanfranchi:
Work started in PR #612 by @atrettin, hopefully this fully fleshes out that idea. Please test from my branch, @atrettin & provide any feedback.

Note that the big change here is to make `is_lin` and `is_log` describe binning behavior, and not anything to do with whether or not the binning is "regular" in linear- or log-space. There is an additional property of `OneDimBinning` called `is_irregular` which will test the regularity of the binning in the space in which it is defined, if that information is useful / necessary (this is used, in part, in the new `__str__` method to allow for more concise string representation of binnings; also could be useful for optimizing certain functions on the bin edges).

## Full run-down of changes:

* `is_lin` and `is_log` no longer imply uniformity in linear-/log-space
* still can use `is_lin` and `is_log` at `__init__`, but if both are none, `is_lin` is set to True (and `is_log` is set to False)
* still use both `is_lin` and `is_log` as internal state (backwards compatibility)
* make `is_log`, `is_lin` setter attributes
* make is_irregular a derived quantity, cached in its getter (and invalidated if setters `is_log` or `is_lin` cause either to be modified)
* update `__init__` handling accordingly
* update docstrings accordingly
* new `__str__` output that always describes lin/log behavior
* rename `is_bin_spacing_lin` -> `is_bin_spacing_lin_uniform`, `is_bin_spacing_log` -> `is_bin_spacing_log_uniform`
* oversample function modified to work bin-by-bin regardless of nature of binning (i.e., works for irregular binning, and doesn't check for regularity)
* oversample bugfix: if factor < 1 error message: ">= 0" -> ">= 1"
* update outdated pylint disables
* disable line length checking for docstring examples
* fix docstring doctest examples; all pass but a few that aren't expected to pass
* fix `basename` function (previously didn't do what it advertised), add more docstring examples  (doctest)

### DIFF
--- a/pisa/core/binning.py
+++ b/pisa/core/binning.py
@@ -206,16 +206,16 @@ class OneDimBinning(object):
     >>> from pisa import ureg
     >>> from pisa.core.binning import OneDimBinning
     >>> ebins = OneDimBinning(name='energy', is_log=True,
-    ...                       num_bins=40, domain=[1, 80]*ureg.GeV)
+    ...                       num_bins=40, domain=[1, 100]*ureg.GeV)
     >>> print(ebins)
-    OneDimBinning('energy', 40 logarithmically-uniform bins spanning [1.0, 80.0] GeV)
+    OneDimBinning('energy', 40 logarithmically-regular bins spanning [1.0, 100.0] GeV (behavior is logarithmic))
     >>> ebins2 = ebins.to('joule')
     >>> print(ebins2)
-    OneDimBinning('energy', 40 logarithmically-uniform bins spanning [1.60217653e-10, 1.281741224e-08] J)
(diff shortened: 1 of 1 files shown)


atrettin:
Thank you for getting all this done so quickly! I will check it out now to see if it works for me.

atrettin:
I'm still having issues loading hypersurfaces because the binning hashes don't match... I'm looking into it.

atrettin:
Okay, I found the problem: In my first version, the `is_log` property for the PID binning with edges `[0.55, 0.75, 1.0]` was inferred to be `True`, when it should be (and now, is) linear.

PR #616, opened 2020-04-22
Adding stage to fix the MC error to initial values.  Needed for mod_chi2

hignight:


pisa/stages/utils/pi_fix_error.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def setup_function(self):
+        for container in self.data:
+            container['frozen_errors'] = np.empty((container.size), dtype=FTYPE)
+
+    def apply_function(self):
+        for container in self.data:
+            vectorizer.assign(vals=container['frozen_errors'], out=container['errors'])
+            container['errors'].mark_changed()
+            container['weights_squared'].mark_changed()
```
philippeller, line 73: seems like this line with "weights_squared" is not needed, as they appear nowhere (?)

pisa/stages/utils/pi_fix_error.py
```diff
@@ -0,0 +1,78 @@
...
+        assert self.output_mode == 'binned'
+
+    def setup_function(self):
+        for container in self.data:
+            container['frozen_errors'] = np.empty((container.size), dtype=FTYPE)
+
+    def apply_function(self):
+        for container in self.data:
+            vectorizer.assign(vals=container['frozen_errors'], out=container['errors'])
+            container['errors'].mark_changed()
```
philippeller, line 72: all `mark_changed` calls should have `WHERE` as an argument, such that this works correctly on GPUs

pisa/stages/utils/pi_fix_error.py
```diff
@@ -0,0 +1,78 @@
...
+
+from pisa import FTYPE
+from pisa.core.pi_stage import PiStage
+from pisa.utils.profiler import profile
+from pisa.utils import vectorizer
+
+
+class pi_fix_error(PiStage):  # pylint: disable=invalid-name
+    """
+    stage to histogram events
```
philippeller, line 20: change docstring accordingly

pisa/stages/utils/pi_fix_error.py
```diff
@@ -0,0 +1,78 @@
...
+            error_method=error_method,
+            input_specs=input_specs,
+            calc_specs=calc_specs,
+            output_specs=output_specs,
+            output_calc_keys=output_calc_keys,
+            input_apply_keys=input_apply_keys,
+            output_apply_keys=output_apply_keys,
+        )
+
+        assert self.input_mode is not None
```
philippeller, line 61: would it work for events in and binned out? If not tested maybe better enforce that to be binned too

atrettin:
Can confirm from testing that the stage works as intended, the errors stay indeed the same after changing parameters in a pipeline.

philippeller:
Great, once the few requested changes are in gonna merge the PR

pisa/stages/utils/pi_fix_error.py
hignight: Fair....that was left over from N tries ago and didn't get removed.

pisa/stages/utils/pi_fix_error.py
hignight: Sorry, didn't know about this.  Imported WHERE and added to all mark_charned

pisa/stages/utils/pi_fix_error.py
hignight: Sorry.  Changed.  Not sure how descriptive to make it....

pisa/stages/utils/pi_fix_error.py
hignight: Fair...I can test later but I will add in the assertion here.  I don't know why it wouldn't though

PR #617, opened 2020-04-26
`PiStage`: allow reasonable inputs; add docstrings

jllanfranchi:


philippeller:
Looks good, only suggestion: consider moving the args to tuple function into pisa/utils/format.py

jllanfranchi:
I had thought about that, thought about the fact that to now this exact behavior is only used in one place, and just put it here. But if we collect all such functions in `utils.format`, then maybe it'll become clear how to merge their functionality (I already have one such function there). So I'll move it there.

Issue #492, opened 2018-04-22
Gaussian random variates from maps ignore `random_state`


thehrh:
Compare https://github.com/jllanfranchi/pisa/blob/dev/pisa/core/map.py#L890 to https://github.com/jllanfranchi/pisa/blob/dev/pisa/core/map.py#L908 and https://github.com/jllanfranchi/pisa/blob/dev/pisa/core/map.py#L931. Looks like an oversight.

PR #618, opened 2020-04-28
use random state for gaussian random variates also

thehrh:
Fix for #492 (also in #587, but I figured this better be fixed before it trips us up)


philippeller:
Looks good to me!

jllanfranchi:
Likewise looks fine to me; checks pass, so I'm merging.

PR #621, opened 2020-05-01
Facilitate upsampling of weighted histograms

atrettin:
This PR adds a stage to resample weighted histograms. It currently only implements upsampling, but downsampling would be a relatively easy extension that someone can make if they need it.

Upsampling is only supported when the bin edges of the fine (output) binning are a superset of the bin edges of the coarse (input) binning. The contents of the upsampled bins are calculated under the assumption that events are distributed uniformly within each coarse bin, meaning that the expectation value in each fine bin is proportional to its fraction of the volume of the coarse from which it was sampled. The variance &sigma;<sup>2</sup> is also scaled by bin volume. I will give a presentation showing why that is the right way to scale errors under the assumptions made.

This is currently just a draft and will need testing. It is also still missing the important feature of calculating bin volumes in log-space, such that the assumption of uniform event distributions is more sensible for bins in energy.


atrettin:
Perhaps it would be better to configure a "variance key" rather than the "error method" to be more explicit and general? As it stands now the stage would break if someone decided to use a key other than `weights_squared` in the `pi_hist` stage...

atrettin:
I have now verified that the stage works and that the math checks out in the end. It is now ready for its original intended purpose. It needs PR  #622  with logarithmic bin volumes to work.

atrettin:
The stage is now compatible with the error fixing stage. I added a line that was missing to create the container for frozen errors in the correct binning.
I tested the entire chain we need now:

weights --> coarse histogram --> hypersurface on coarse bins --> upsample --> fix errors

So this should be all we need to get our analysis working with this.

atrettin:
Converted back to draft because there are still issues with containers during minimization that need to be resolved.

philippeller:
This looks good to me!

Issue #619, opened 2020-04-30
Binning enhancement to facilitate resampling


atrettin:
I'm currently working on the resampling stage and there are two enhancements to the binning classes that would make it more practical:

1. Rename `is_compat` to `is_downsample` or something to the effect. That would make it clearer what the function is testing for. A stage can just exchange the order of the comparison to check for upsampling compatibility. This also allows for automatic switching between upsampling and downsampling modes in a more generalized resample stage.
2. Make it possible to get bin volumes for bins in log dimensions in "units" of orders of magnitude. In other words, make bins that have the same size visually in a plot have the same volume. The assumption that events are evenly distributed over bins that are being upsampled into makes more sense in log-energy than in energy.

jllanfranchi:
If you read the comments and TODO's in the sourcecode around `is_compat`, you might have gleaned that I originally imagined this would be a function to test whether one could _either_ upsample _or_ downsample from the binning of `self` to the binning of `other`. But that never became a use case, and you are now using it in a test that requires that it return True if one can downsample, and False even if one can upsample. That's obviously a valid thing to want to do, but I agree that either the name needs to change. To make it "grammatical," maybe:
* `can_upsample_to`: can binning of `self` be upsampled to binning of `other`
* `can_downsample_to`: vice versa
* `can_resample_to`: either way

There's further complication in the terminology and the assumptions that we're making in that arbitrarily re-binning of histograms without requiring that the new bin edges are a super-/subset of the source binning shouldn't be terribly difficult and is likely the next step someone will want to make from here. From that point of view, only a check that dimensionality is equivalent is necessary (and maybe one would optionally want to have a flag that checks that no extrapolation is necessary to get from one binning to the other... or that's the default check, and the flag allows for extrapolation).

Maybe this simple kind of dimensionality check, that accommodates arbitrary re-binning, is the kind of thing that we place in `is_compat`.

In fact, now that I'm thinking about it, if you have the logic and/or implementation of upsampling already, then I don't see why (so long as all edges are within the domain of the original binning) you can't implement resampling with arbitrary bin edges. E.g., if you start with binning (in 1d, the bookkeeping gets a bit more hairy in higher dimensions)
```
|______|______|
```
and you want to go to binning
```
 |___|___|___|
```
then you take the union of the two binnings:
```
 |___|_|_|___|
```
[...]

PR #622, opened 2020-05-01
Add weighted_bin_{widths,volumes}; bugfix

jllanfranchi:
* Add `OneDimBinning.weighted_bin_widths`, `MultiDimBinning.weighted_bin_volumes`
* Allow `meshgrid` method of `MultiDimBinning` to accept the name of any attribute, allow failure
* Bugfix: invalidate cached items if is_{log,lin} changes

As of now, this is largely untested. I'll let @atrettin start playing with it (maybe you can even add a unit test or two?) and then we can merge this PR and create subsequent PR's to add more features to help the rebinning effort (if this is useful already), or I'll add commits to this PR with the other features requested as I develop them. (Let me know what's preferable.)

Note this PR addresses point (2) of Issue #619 (Binning enhancement to facilitate resampling), ultimately in support of PR #621 (Facilitate upsampling of weighted histograms)

### DIFF
--- a/pisa/core/binning.py
+++ b/pisa/core/binning.py
@@ -130,7 +130,7 @@ def _new_obj(original_function):
         """<< docstring will be inherited from wrapped function >>"""
         new_state = OrderedDict()
         state_updates = original_function(cls, *args, **kwargs)
-        for attr in cls._hash_attrs: # pylint: disable=protected-access
+        for attr in cls._attrs_to_create_new:  # pylint: disable=protected-access
             if attr in state_updates:
                 new_state[attr] = state_updates[attr]
             else:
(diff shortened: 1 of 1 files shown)


atrettin:
Wait, hoooold up! I think we have a problem... the sums of bin volumes of two binnings don't match anymore when they are in log and one is a strict down-sample of the other! That was a bit of an oversight...

pisa/core/binning.py
```diff
@@ -956,6 +971,18 @@ def bin_widths(self):
             self._bin_widths = np.abs(np.diff(self.bin_edges.m)) * self.units
         return self._bin_widths
 
+    @property
+    def weighted_bin_widths(self):
+        """Absolute widths of bins."""
+        if self._weighted_bin_widths is None:
+            if self.is_log:
+                self._weighted_bin_widths = (
+                    self.edge_magnitudes[1:] / self.edge_magnitudes[:-1]
```
atrettin, line 980: I think we have to take the log of this to make the sums of two binnings add up again!

atrettin: After testing my stage again I can confirm that taking the log of the ratio here produces the correct total event rates after upsampling. If we make this log, I would suggest also placing a comment in the code telling people why it has to be this way so that no one accidentally breaks the upsampling stage.

jllanfranchi: I can add a comment, but a comment _and_ a unit test would be even better! Can you checkout from my branch here, make the modifications you propose, add an appropriate unit test that catches this issue (so we have more than just the comment in the code to check us -- though I think that's a good idea, too!), and submit a PR against my branch? I can merge there, then merge the whole thing into PISA. (I've got a lot going on right now, so you doing this will help move things along faster.)

atrettin: Ok yes, I can do that, but for this evening I think I have coded myself out... I can get to it tomorrow!

atrettin:
Alright, all tests passed, it looks like we're good to go!

PR #623, opened 2020-05-07
add options for labeling bins

atrettin:
This PR just adds some more options for labeling bin counts when plotting a Map. You can automatically switch colors for better contrast and strip leading zeros. Existing options remain unchanged.


jllanfranchi:
Looks good and non-obtrusive, while adding features. Will merge presently.

PR #625, opened 2020-05-18
I3open master

bourdeet:
This PR changes the way events are reweighted in cases where the axial mass parameter falls below the range where points where provided by GENIE. In some cases, the polynomial interpolation used to fit the reweighting curve over the range [-2,2] sigma led to negative weights. The proposed fix is to clip these negative events and floor their values to 0.0

I also added a warning in the setup_function, notifying the user if their defined fit range is larger than [-2,2]


pisa/stages/xsec/genie_sys.py
```diff
@@ -113,9 +126,17 @@ def apply_function(self):
                 container['quad_fit_maccres'].get(WHERE),
                 out=container['weights'].get(WHERE),
             )
+
+            #
+            # In cases where the axial mass is extrapolated outside
+            # the range of the points used in the interpolation, some 
+            # weights become negative. These are floored at 0.
+            #
+            container['weights']= np.clip(container['weights'].get(WHERE),a_min=0.,a_max=np.inf)
```
philippeller, line 135: this is a CPU only operation, so you should get `host` arrays

PR #626, opened 2020-05-19
Merging oscNext high stats sample PISA branch

ts4051:
A few changes implemented as part of the oscNext high stats sample analyses.

Description of changes:
1) Data loading changes, contained in `simple_data_loader`, `events_pi` and `hdf` (see dedicated PR #607, this is the same changes so if we merge that one first this one is then handled too)
2) Fix to theta23 octant handling (was a bug where the forcing into differnt fit octants didnt work for a particular case)
3) Made fitting octants separately the default behaviour (this seems to be the consensus on what the best practice is)
4) Fix for case where user's numba version doesn't have the `NumbaDeprecationWarning`
5) User can specify `output_mode` when calling `Pipeline.get_outputs()` if they want to request e.g. events instead of binned data. Useful for making plots from the output of the PISA pipeline.
6) Improvements to MCEq stage speed, and a few general style improvements to the code


pisa/__init__.py
```diff
@@ -11,7 +11,14 @@
 import warnings
 
 from numba import jit as numba_jit
-from numba import NumbaDeprecationWarning
+
```
jllanfranchi, line 14: All of these changes should be unnecessary because we require numba 0.45.1. That said, you can do the following (and get rid of all the "FOUND..." stuff):
```python
try:
    # numba 0.45.1
    from numba.errors import NumbaDeprecationWarning
except ImportError:
    # latest numba versions
    from numba.core.errors import NumbaDeprecationWarning
```
and that should be compatible with Numba versions newer than 0.45.1 (of course `SmartArray`s go away)

pisa/stages/flux/pi_mceq_barr.py
```diff
@@ -499,9 +500,19 @@ def apply_sys_kernel(
...
+
+    # Nominal flux + spectral index change
+    result = nu_flux_nominal * spectral_index_scale(true_energy, energy_pivot, delta_index)
+
+    # Apply bar params
+    result += np.dot(gradients, gradient_params)
+
+    # Check for negative results from spline (np.clip not supported by vectorization)
+    negative_mask = result < 0
+    if np.sum(negative_mask):
```
jllanfranchi, line 512: 1. I'm guessing you mean to do `result[negative_mask] = 0.0` on L513?
2. Is the `if` clause on L512 premature / assumed optimization? Because it looks like the opposite to me. In the time you take to check every value in the mask (and you are even doing more than that; you're performing a _sum_ over the entire mask!), you might as well have just applied the mask.

So, unless you have actual timing tests run, I'm guessing it will be faster--and I know it removes LOC--if you simply do
```python
result[result < 0.0] = 0.0
```

pisa/stages/flux/pi_mceq_barr.py
```diff
@@ -25,6 +25,21 @@
 from pisa.utils.resources import find_resource
 
 
+def eval_spline(true_log_energy, true_abs_coszen, spline):
```
jllanfranchi, line 28: This looks like a redundant function; please remove, unless there's a good reason for it that I am missing.

pisa/core/pi_stage.py
```diff
@@ -287,20 +288,32 @@ def run(self, inputs=None):
...
         event container itself
+
+        Parameters
+        ----------
+        output_mode: None, string
+            Optionally can force the output mode to the stage
         """
+
+        # Handle optional user-overridden `output_mode`
+        if output_mode is None :
```
jllanfranchi, line 304: no space before colons (another on L306)

pisa/stages/flux/pi_mceq_barr.py
ts4051: Fixed

pisa/core/pi_stage.py
ts4051: Fixed

pisa/stages/flux/pi_mceq_barr.py
ts4051: Function removed. I think this function used to do more than this but now it is a one-liner yes can just directly call that one line in the code.

pisa/__init__.py
ts4051: I am using numba 0.43.1 which is where I have the error, and it looks like `NumbaDeprecationWarning` doens't exist there. I remember there was some issue with the default numba which is why I went to an older one but can't remember now what tat was (might have been a llvmlite version issue). But sounds like this is specific to non-standard stuff in my install then, so probably best just to drop this change.

ts4051:
Thanks for the review. All suggestions have been implemented now. Let me know if anything else comes, particularly when you test drive the code.

pisa/analysis/analysis.py
```diff
@@ -266,17 +266,22 @@ def get_separate_t23_octant_params(hypo_maker,inflection_point) :
...
-    tolerance = 1. * ureg.degree
-    if np.isclose(theta23.value.m_as("degree"),45.,atol=tolerance.m_as("degree")) : 
-        theta23.value -= tolerance
-
+    # If theta23 is maximal (e.g. the transition between octants) or very close 
+    # to it, offset it slightly to be clearly in one octant (note that fit can 
+    # still move the value back to maximal). The reason for this is that 
+    # otherwise checks on the parameter bounds (which include a margin for 
+    # minimizer tolerance) can an throw exception.
+    tolerance = 0.1 * ureg.degree
```
jllanfranchi, line 278: Why wouldn't a value within `tolerance` of maximal simply be placed at `tolerance` away from maximal, rather than having `tolerance` added (subtracted) from the value?

E.g., if a user (or whatever sets it) asks for t23=44.901 deg, doesn't setting t23=44.9 deg seem like the better "approximation" to what the user wants than 44.801 deg?

I don't perceive any benefit to having a "linear" shifting behavior for this operation. But if I'm missing something, please let me know.

thehrh: What do you think of the solution in my PR? https://github.com/IceCubeOpenSource/pisa/pull/587/files#diff-5f6e672ea43f8f20cf02b7c62bafe16fR151 (and following)

ts4051: Hi Justin, I think what your suggesting the code should be doing is what it is doing. E.g. If you are closer than `tolerance` to 45 degrees, you are moved to 45 +/- tolerance (where the sign is chosen to stay in the same octant).

ts4051: It looks to me like Thomas' solution is doing the same thing. So either implementation is fine by me.

jllanfranchi: You're right, Tom; I misread what the code was doing. So we're good on this.

PR #627, opened 2020-05-20
Introducing new output format to binned data + new generalized poisson likelihood

bourdeet:
This PR contains all that is needed to run the generalized poisson likelihood. There are a lot of changes affecting many functions, some of which coming from the fact that I based this branch off the oscNext 2 branch that is currently being merged in PR#626. I'll make sure to merge the outcome of that PR later, but for the moment, I can still introduce the key changes that are required for the likelihood to work.

NOTE: for the likelihood to work, one must compile a set of Cython-compiled functions which is not in our current standard list of programs automatically installed in pip. to compile the functions, go in the folder `llh_defs`, and call "python setup.py build_ext --inplace"

 **A: changes to pi_stage.py**
---
The first things that was needed was a way to obtain maps of quantities other than `weights` when I ran the `get_output` function of `pi_stage`. The way I chose to do this was to allow for the function to return dictionaries of Mapsets, where you would get fx. D['weights'], D['quantity a'], D['quantity b'] for every quantity contained in the stage's `output_apply_key`

I added an option which by default returns the normal object expected by usual pisa analyses

The changes made to pi_stages then had to be propagated to `pipeline.py` and `distribution_maker.py`

**B: New modules to precompute quantities for the likelihood`**
---
**`add_indices.py`**
the generalized llh needs information about the weight distribution in each bin. the module `add_indices` maps out every individual events to the bin in which it falls to that end. Since MC events are not migrating between bins in our current pipeline (since we don't touch reconstructed energy and zenith and pid during the minimization), this stage can be run pretty much anywhere in the pipeline.

**`prepare_generalized_llh_parameters.py`**
In order to speed up calculations of the likelihood during optimization, I chose to compute each bin's statistical moments in a module called `prepare_generalized_llh_parameters.py`. This computes the `alphas` and `betas` than are used to approximate the gamma-shaped uncertainty of a bin's weight distribution. That function also propagates the hypersurface error on individual events.

**C The likelihood function itself**
---
I added the likelihood computation in the `stats.py` file, and likewise added its method in the `Map` class which is where the function is called during an analysis. A persisting problem I have right now is that the dict format of the output (as described in A) prevented me from calling the metric directly from a MapSet object, as it is done with the other optimization metric. If anyone has an idea on how to fix this, I would appreciate tips!

**D changes to the Analysis Class**
---
A couple of changes had to be made there to allow for the `fit_inner_hypo` and `_minimizer_callback` functions of the `Analysis` to handle the new output format (dict instead of MapSets), the use of kwargs when calling metrics, and save the metric info after minimizing. I think that my footprint there could be attenuated by fixing this Mapset issue mentioned in C. The later stages that follow successful minimization might need a bit of work there, as I didn't really care about the metric info from the generalized poisson lh too much.

**E A neat little toy MC pipeline to test all of this
---
the `super_simple_signal.py` stage is a simple stage that produces a gaussian signal on top of a random and uniform background. the associated pipeline config file is `supersimple/ super_simple_pipeline.cfg`, and an example script that runs the plots I showed in [ this talk](https://drive.google.com/file/d/1bLaHT8zEgYR2NJm47S6kXqzgdUzpqEa5/view) is included and named `likelihood_1D_test.py`

**F everything else that's changed
---
these include changes that are currently being reviewed in PR #626, as well as a bunch of function from thorsten's [mc_uncertainty git](https://github.com/thoglu/mc_uncertainty). I am in the process of cleaning these so don't worry about these files for now

I hope these explanations help a bit :)


bourdeet:
Hey!
Yes, I would like the key functions of `test_bourdeet` to be included into the core of pisa. I kept all my written stage in the same folder for now, but I assumed they would eventually be moved into whatever location is deemed appropriate.

For now, I'll leave the c files in so that people who want to test out my module can do so by compiling the minimum dependencies required.

bourdeet:
Ok I think I've cleaned the PR of all the files I didn't need. I still need to fix / move a couple of the test scripts to the fridge, but those are not the most important items.


I would like you guys to focus on the changes I have in `pi_stage.py`, which is the key change that affects the core pisa code. I want to know if formatting the outputs into dict of multiple mapset is something you agree on, and if not, I want to know how I can change this to be able to still make the likelihood implementation work

pisa/analysis/analysis.py
```diff
@@ -762,6 +798,8 @@ def fit_hypo_inner(self, data_dist, hypo_maker, metric, minimizer_settings,
 
         # Record the Asimov distribution with the optimal param values
         hypo_asimov_dist = hypo_maker.get_outputs(return_sum=True)
+        gen_poisson_dist = hypo_maker.get_outputs(return_sum=False, force_standard_output=False)
```
philippeller, line 801: what does "gen" stand for? My immediate guess would be "generated".....but knowing your PR it's probably "generalized"? no need to abbreviate.....I'd rather have longer names that are clear

philippeller: Also, are these outputs now always retrieved, even if I fit a simple TS? (I'm a bit concerned about speed here)

pisa/core/distribution_maker.py
```diff
@@ -275,6 +292,43 @@ def source_code_hash(self):
...
+        '''
+        num_bins = self.pipelines[0].stages[-1].output_specs.tot_num_bins
+        self.num_mc_events_per_bin = np.zeros(num_bins)
+
+        for p in self.pipelines:
+
+            for c in p.stages[-1].data:
+
+                # skip data or pseudo-data pipelines
+                if c.name == 'data' or c.name == 'Total':
```
philippeller, line 311: is this a good idea to treat data, or what you think is pesudo data because the name is "Total" any different?
I'd rather not make this distinction here. And for (pseudo)data you're anyway only interested in expected values.....so having extra info won't hurt.

What if i have a pipeline of my model called "Total".....your code will then unexpectedly not work. Or what if my pseudodata is called "total" and not "Total", then behaviour will be different

pisa/utils/stats.py
```diff
@@ -127,8 +127,6 @@ def chi2(actual_values, expected_values):
         #       still destroy a minimizer's hopes and dreams...
 
         # Replace 0's with small positive numbers to avoid inf in division
-        np.clip(actual_values, a_min=SMALL_POS, a_max=np.inf,
```
philippeller, line 130: why are all the clipping of values removed??

philippeller:
* Also check there are unit tests failing
* And cam the compilation of your llh function be done in the regular setup.py?

pisa/stages/test_bourdeet/add_indices.py
```diff
@@ -0,0 +1,136 @@
...
+        Create one mask for each analysis bin.
+        '''
+        
+        assert self.calc_specs == 'events', 'ERROR: calc specs must be set to "events for this module'
+
+        self.data.data_specs = 'events'
+
+        for container in self.data:
+            # Generate a new container called bin_indices
+            container['bin_indices'] = np.empty((container.size), dtype=FTYPE)
```
jllanfranchi, line 118: shouldn't indices be an integer type?

pisa/stages/test_bourdeet/add_indices.py
```diff
@@ -0,0 +1,136 @@
...
+  
+            e = container['reco_energy']
```
jllanfranchi, line 120: you should make this adapt to the binning; this is hard-coded to work only with specific binning spec (even changing the order will break this code). (I believe output specs should be able to tell you what you need to do this dynamically.)

bourdeet:
just answering a few questions that were lingering from phillip:

- The compilation of the llh function is now integrated into the setup.py script of pisa
- I've cleaned up the scripts and errors that gave problems for the unit test. the latter have worked fine in the latest commit
- I have written down the full name of the variable (generalized instead of gen), so there won't be any confusion
- The new output format is given only is the metric reuested is the generalized likelihood. it is now calculated in an if loop, so no worries about the memory here.

pisa/core/distribution_maker.py
philippeller: Hi Etienne, test test

bourdeet: testing on my end!

bourdeet: wakawaka

pisa/stages/test_bourdeet/add_indices.py
bourdeet: I've modified that line to be more generic now

pisa/utils/stats.py
philippeller: aha, good point! in the old days we used the Neyman chi2 dividing by observed, but now in the Pearson chi2 this is not the case....so as you point out we can leave the actual values alone

philippeller:
My comments have been addressed, so from my side we can get this merged

bourdeet:
cool! I'll just change the name of the paths before things get merged

PR #628, opened 2020-05-25
enable deletion of Param from ParamSet

atrettin:
Does what it says on the tin: You can now delete items from a ParamSet. Comes with unit tests to make sure it works as advertised.


jllanfranchi:
Looks good to me! You inspired me, I have a subsequent PR that will make this fully implement the interface of MutableSequence, plus some goodies from Set.

PR #630, opened 2020-05-26
Explicit check on osc angle param units

ts4051:
There is a bug case where if an oscillation angular param (e.g. mixing angle of CP phase) is dimensionless, it is interpreted as radians in the osc stage. This is a problem is the user meant degrees but forgot the unit. This check enforces the user specified the unit.


jllanfranchi:
First, please rebase from i3os/master (or start there when you create a feature branch) so you are up to date with i3os/master, the commit diffs appear reasonable, and github isn't fooled into digging up old PR conversations.

Second, you can implement a better check than this using the `dimensionality` attribute.  E.g. 
```python
angle_dim = ureg.deg.dimensionality
for blah in blah:
    assert x.dimensionality == angle_dim
```

Third, check != and then raise `pint.DimensionalityError` (instead of using `assert`). The more typed an exception can be the better, and there's a perfect exception type that can be used for this case.

philippeller:
@ts4051 will you make the remaining small changes and rebase? Then we can get this thing merged

philippeller:
@ts4051 any updates on the last changes and rebase? Would be good to get this merged, seems 99% there

philippeller:
looks good now i think...gonna merge

PR #633, opened 2020-07-10
Negative muon weight

bourdeet:
Once again, Fixing instance f a stage producing individual negative weight.


bourdeet:
turns out I was just behind the master by a few commits. Sorry about that!

PR #636, opened 2020-09-08
introduce two blindness levels, fix bug in highest blindness level

atrettin:
This PR facilitates the blind fit procedure we are planning to follow (at least for the verification sample). When the `blind` flag evaluates to `True`, fit parameters are prevented from being printed to the console or showing up in logs, but the fitted parameters are still stored in the output dictionary. If an int > 1 is passed to `blind`, the parameters are also prevented from being stored in the output. That means `blind == 2` behaves as the previous implementation while `blind == 1` or `blind == True` is a little more relaxed. The way the fridge scripts are currently set up, the final fit results are stored in compressed JSON files that are not human readable to maintain blindness, but scripts can use them to produce post-fit data/MC comparison plots.

This PR also fixes a bug where the fit would crash after checking the second octant separately due to the way the hypo_maker is reset after the inner loop.

### DIFF
--- a/pisa/analysis/analysis.py
+++ b/pisa/analysis/analysis.py
@@ -428,10 +428,12 @@ class Analysis(object):
         pprint : bool
             Whether to show live-update of minimizer progress.
 
-        blind : bool
-            Whether to carry out a blind analysis. This hides actual parameter
-            values from display and disallows these (as well as Jacobian,
-            Hessian, etc.) from ending up in logfiles.
+        blind : bool or int
+            Whether to carry out a blind analysis. If True or 1, this hides actual
+            parameter values from display and disallows these (as well as Jacobian,
+            Hessian, etc.) from ending up in logfiles. If given an integer > 1, the
+            fitted parameters are also prevented from being stored in fit info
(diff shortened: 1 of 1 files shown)


pisa/analysis/analysis.py
```diff
@@ -808,8 +812,10 @@ def fit_hypo_inner(self, data_dist, hypo_maker, metric, minimizer_settings,
         fit_info = OrderedDict()
         fit_info['metric'] = metric
         fit_info['metric_val'] = metric_val
-        if blind:
-            hypo_maker.reset_free()
+        if blind > 1:  # only at stricter blindness level
```
philippeller, line 815: just out of curiosity: why is this only needed during blind fits but not otherwise?

atrettin: At the strictest blindness level, no information is stored about the fits. This reset command overwrites any fitted values to make sure they can't be stored even accidentally later.

philippeller: Aha, ok got it! Sounds good. I'll merge this PR then

PR #637, opened 2020-09-12
Tiny llh fixes

bourdeet:
This PR makes a series of name changes to remove the prototype nomenclature, and make the stages more compliant with the pisa conventions.


pisa/stages/simple_fit/pi_simple_signal.py
```diff
@@ -16,7 +16,7 @@
 from pisa.core.bin_indexing import lookup_indices
```
philippeller, line 16: It might be better to move this to `/stages/data` as that's the place where we have other dummy data stages

pisa/stages/likelihood/llh_defs/poisson.py
```diff
@@ -31,7 +31,7 @@
 import numpy as np
```
philippeller, line 31: maybe should me move all of those into `pisa.utils`?

philippeller (commented):
Hi,

In general looking good, I just added two suggestions about file locations

pisa/stages/simple_fit/pi_simple_signal.py
bourdeet: sounds good, though I noticed that the README documentation about this folder is quite out of date. Do you mind if I make a few changes there too?

philippeller: Be absolutely my guest!

PR #639, opened 2020-09-19
Osc layers

bourdeet:
Looking a bit more closely into this stage, I noticed that path lengths seem to be calculated in a way that is not very straightforward to understand. This PR tries to simplify the computations made in layers.py to make them easier to understand. I also added a few assert statements to prevent one from defining unphysical detector locations and radii definitions.

Note that the only thing that has changed computationaly is the way the full path length, as computed in extCalcLayers and calcPathLength. The logic and steps behind the multi layer calculations should probably also be looked at, though it will have to be in a later PR :)

### DIFF
--- a/pisa/stages/osc/layers.py
+++ b/pisa/stages/osc/layers.py
@@ -126,8 +126,8 @@ def extCalcLayers(cz,
             calculate_small_root = (coszen < coszen_limit) * (coszen_limit <= coszen_limit[idx])
             calculate_large_root = (coszen_limit>coszen)
 
-            small_roots = - r_detector * coszen.astype(FTYPE) * calculate_small_root.astype(FTYPE) - np.sqrt(r_detector**2 * coszen.astype(FTYPE)**2 - r_detector**2 + radii.astype(FTYPE)**2) #, where=calculate_small_root, out=np.zeros_like(radii))
-            large_roots = - r_detector * coszen.astype(FTYPE) * calculate_large_root.astype(FTYPE) + np.sqrt(r_detector**2 * coszen.astype(FTYPE)**2 - r_detector**2 + radii.astype(FTYPE)**2) #, where=calculate_large_root, out=np.zeros_like(radii))
+            small_roots = - r_detector * coszen * calculate_small_root.astype(FTYPE) - np.sqrt(r_detector**2 * coszen**2 - r_detector**2 + radii.astype(FTYPE)**2) #, where=calculate_small_root, out=np.zeros_like(radii))
+            large_roots = - r_detector * coszen * calculate_large_root.astype(FTYPE) + np.sqrt(r_detector**2 * coszen**2 - r_detector**2 + radii.astype(FTYPE)**2) #, where=calculate_large_root, out=np.zeros_like(radii))
 
             # Remove the negative root numbers, and the initial zeros distances
             small_roots = small_roots[small_roots>0]
--- a/pisa/stages/osc/layers.py
+++ b/pisa/stages/osc/layers.py
@@ -126,6 +126,11 @@ def extCalcLayers(cz,
             calculate_small_root = (coszen < coszen_limit) * (coszen_limit <= coszen_limit[idx])
             calculate_large_root = (coszen_limit>coszen)
 
+            logging.info('r_detector: {}'.format(type(r_detector)))
+            logging.info('coszen: {}'.format(type(coszen)))
+            logging.info('radii: {}'.format(type(radii)))
+            logging.info('calculate: {}'.format(type(calculate_small_root)))
+
             small_roots = - r_detector * coszen * calculate_small_root - np.sqrt(r_detector**2 * coszen**2 - r_detector**2 + radii**2) #, where=calculate_small_root, out=np.zeros_like(radii))
             large_roots = - r_detector * coszen * calculate_large_root + np.sqrt(r_detector**2 * coszen**2 - r_detector**2 + radii**2) #, where=calculate_large_root, out=np.zeros_like(radii))
 
--- a/pisa/stages/osc/layers.py
+++ b/pisa/stages/osc/layers.py
@@ -113,8 +113,8 @@ def extCalcLayers(cz,
             # the max number of layers
             n_zeros_pads = max_layers - segments_lengths.shape[0]
 
-            segments_lengths = np.concatenate((np.zeros(n_zeros_pads, dtype=FTYPE), segments_lengths))
-            density = np.concatenate((np.zeros(n_zeros_pads, dtype=FTYPE), density))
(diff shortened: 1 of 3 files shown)


pisa/stages/osc/layers.py
```diff
@@ -101,14 +101,11 @@ def extCalcLayers(
...
         # Above horizon
         if coszen >= 0:
-            kappa = (detector_depth + prop_height)/r_detector
-            path_len = (
-                r_detector * np.sqrt(coszen**2 - 1 + (1 + kappa)**2)
-                - r_detector * coszen
-            )
-
             # Path through the air:
             kappa = detector_depth / r_detector
```
bourdeet, line 110: This line was particularly dangerous, as it redefined the variable kappa mid- condition. Taking the total path length out of this loop gets rid of this issue

philippeller:
Very nice!
Can I ask for one more thing here before we merge? We should make the test function at the end of this file stronger, i.e. calculate some layers and then actually assert that they are what we expect. So it will prevent future corruption of this function.
Use `np.allclose` with the PISA allclose kw arg for that

bourdeet:
Okay! I started writing some tests, and I can already see that there seem to be an error in the way layer distances are computed as of now. It's quite cumbersome and difficult to properly describe this problem in words, so I've made a couple of drawings to reprent things. See this PDF linked here: https://icecube.wisc.edu/~bourdeet/oscnext_technote_plots/layer_debugging/layer_bug_hunt.pdf

A double check on the math here would be appreciated!

Tomorrow, I'll dig into the other half of the problem...

bourdeet:
Quick update on the second part of the code: handling paths that cross one or more inner layer:

I have now figured out the math for this section, and once again, there seem to be a bug in the current code. This bug is easily identifiable when considering the case where the path is straight upgoing (ie coszen = -1).

Say that we use the PREM 4 layer model as a reference. That model has layers with the following radii:

earth surface: 6371 km
layer 1:             5701 km
layer 2:            3480 km
layer 3:            1220 km
earth center:            0 km


Starting from a production height of 20 km and a detector depth of 2km, a neutrino would have the following segments along its travel:
[20, 670, 2221, 2260, 1220, 1220, 2260, 2221, 668] 
in the following medium:
 [atmo, surface-to-L1, L1-to-L2, L2-to-L3, L3-to-center, center-to-L3, L3-to-L2, L2-to-L1, L1-to-detector]

Now, if I run extCalcLayers for this direction, the segments returned by the function are the following:

[20, 2., 670, 2221, 2260, 2440, 2260, 2221, 670]
corresponding to travel in the following:
[atmo-to-surf, surf-to-detector, surf-to-L1, L1-to-L2, L2-to-L3, L3-to-L3, L3-to-L2, L2-to-L1, L1-to-surf]

as can be seen here, the current code doesn't include the detector depth properly.

pisa/stages/osc/layers.py
```diff
@@ -67,17 +61,13 @@ def extCalcLayers(
...
-    prop_height
-    detector_depth
-    max_layers
-    min_detector_depth
-    rhos
-    YeFrac
-    YeOuterRadius
-    default_elec_frac
-    coszen_limit
+    cz             : coszen value
```
jllanfranchi, line 64: this is not numpy docstring format

pisa/stages/osc/layers.py
```diff
@@ -422,7 +413,92 @@ def test_Layers():
 
     logging.info('<< PASS : test_Layers >>')
 
+def test_layers_II():
```
jllanfranchi, line 416: Arabic numerals >> Roman numerals; feel free to rename `test_layers` to `test_layers_1` if you want `test_layers_2` to follow some sort of convention

jllanfranchi (commented):
just a couple of small things to note; i haven't checked accuracy yet

philippeller:
we should also get this one merged soon

philippeller:
@bourdeet did you see there was an error when the tests were run in the layers module?

bourdeet:
This one has given me some vectorization problems I have not been able to solve. Help is needed here!

atrettin:
> This one has given me some vectorization problems I have not been able to solve. Help is needed here!

One thing I learned only recently about numba is that, when you are using `jit` with `nopython`, using plain for-loops is actually a _good_ thing! I had so much trouble vectorizing functions because things like `np.dot` didn't work properly and I tried to somehow work around it and it never worked well. All those headaches are actually not necessary, because compiled for-loops are just as fast in numba! I know it's hard to throw all you learned about vectorization with numpy broadcasting over board, but do give it a try, it will make everything much easier. :)

philippeller:
hey Etienne...can you maybe just describe what the operations is supposed to do? I see you reverse, and to diffs and reverse again etc...
Maybe a) there is an easier way to achieve the same. Or then b), as Alex suggests, we can just do this as a for loop

philippeller:
@bourdeet any chance to get this fixed and merged soon? Would be great to do that last PR before rolling out pisa4

pisa/stages/osc/layers.py
```diff
@@ -47,170 +46,141 @@ def __call__(self, *args):
...
+        # Determine if there will be a crossing of layer
+        # idx is the index of the first inner layer
+        idx = np.where(radii<r_detector)[0][0]
+
+        # Deal with paths that do not have tangeants
+        if coszen >= coszen_limit[idx]: 
+            cumulative_distances = -r_detector * coszen + np.sqrt(r_detector**2. * coszen**2. - r_detector**2. + radii[:idx]**2.)
+            # a bit of flippy business is done here to order terms
+            # such that numpy diff can work
+            segments_lengths = np.diff(np.concatenate((np.array([0.], dtype=FTYPE), cumulative_distances[::-1])))
```
atrettin, line 106: It is a bit unfortunate that all of this is done with several concatenations, because that is a very slow operation where the entire array has to be copied. Could you rewrite this with for-loops? I think that it would make the whole thing also easier to understand because you wouldn't have to do all these inversions and concatenations, and will probably solve your ftype problems as well.

bourdeet: For now I believe that the most urgent issue is to get a PR that works merged, so that Philipp can move PISA4 onward. Any issues regarding performances can be addressed by a new PR in the new version of PISA.

philippeller: Agreed!

atrettin: Okay, I made a PR to this PR, @bourdeet ! It fixes the FTYPE issue and all the unit tests pass, so this should be all good now I hope. :)

bourdeet:
alright, all checks passed. hurray!

bourdeet:
ready to merge and leave pisa 3 behind for good :p

PR #641, opened 2020-09-23
pisa4 dev

philippeller:


atrettin:
Whoaaa, the deletions! I'm happy to see PISA hitting the treadmill to lose weight. BTW, if you don't want a PR to be merged, you can mark it as draft. It's a tiny link in the right column iirc.

philippeller:
Cool, marked it as draft!

There is more simplification to come....

PR #642, opened 2020-09-25
Adhoc systematic for additional blind fit checks

atrettin:
One more test to run is to check for those variables in which we see disagreement after the fit, how much our fit would be influenced if we had some systematic to account for it. 

This stage simply rescales all events by the ratio of data/MC in the same binning that was used to make our data/MC comparisons. Those ratios are produced externally and stored to a JSON that is loaded by the new stage. The JSON should simply be a dictionary, where the keys are the variables to be checked and for each key it contains the binning and the data/MC ratios. Our verification sample data/MC comparison script produces these: 

https://github.com/IceCubeOpenSource/fridge/blob/master/analysis/oscnext_verification_sample/post_blind_fit/post_fit_data_mc.py

This also required implementing the 1D lookup function to do the scaling.

### DIFF
--- a/pisa/core/translation.py
+++ b/pisa/core/translation.py
@@ -370,16 +370,23 @@ def lookup(sample, flat_hist, binning):
 
     Notes
     -----
-    Only handles 2d and 3d right now
+    Handles up to 3D.
 
     """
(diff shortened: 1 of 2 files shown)


philippeller:
Looks straight forward enough to me! I can merge this

PR #643, opened 2020-10-09
correcting flaws in the likelihood test script

bourdeet:
This PR corrects a major flaw in the way I originally set up my simple 1D likelihood test. 

Initially, the pipeline I created in pi_simple_signal would generate, at every iteration of the minimizer, a completely new distribution of weights that were randomly distributed according to a gaussian signal. This meant that the actual locations of the weight within the analysis binning varied from iteration to iteration, which is not exactly what I intended to do.

Instead of doing the above, a single distribution of weights is now generated only once, at the setup phase of the stage. These weights are distributed uniformly over the chosen range of parameter values, and during a fit, a gaussian PDF is used to _re-weight_ the already generated weights. This is closer to the scenario we have in real life, as the location of weights in the analysis binning is fixed by the event reconstruction we get out of the simulation production.

This change has the added benefit of making the likelihoods_1D_test.py run much faster, and it also makes the code in pi_simple_signal a bit simpler.

Finally, I've also added two flags to pi_generalized_llh_params, which enable the user to turn on or off the use of pseudo-weights and mean adjustment in the generalized likelihood. When these two flags are set to False, the stage will no modify the input weights.

### DIFF
--- a/pisa/stages/data/super_simple_pipeline.cfg
+++ b/pisa/stages/data/super_simple_pipeline.cfg
@@ -32,7 +32,9 @@ input_specs = None
 calc_specs = None
 output_specs = super_simple_binning
 
-
+# Arguments to turn pseudo weight and mean adjustment on and off
+with_pseudo_weight = False
+with_mean_adjust = False
 
 [pipeline]
 
--- a/pisa/analysis/analysis.py
+++ b/pisa/analysis/analysis.py
@@ -851,7 +851,7 @@ class Analysis(object):
             if blind:
                 msg = ''
             else:
-                msg = ' ' + optimize_result.message
+                msg = ' ' + str(optimize_result.message)
             raise ValueError('Optimization failed.' + msg)
 
         return fit_info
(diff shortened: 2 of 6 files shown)


pisa/stages/data/super_simple_pipeline.cfg
```diff
@@ -32,7 +32,9 @@ input_specs = None
 calc_specs = None
 output_specs = super_simple_binning
 
-
+# Arguments to turn pseudo weight and mean adjustment on and off
+with_pseudo_weight = False
+with_mean_adjust = False
 
 [pipeline]
```
philippeller, line 40: We haven't had cfgs inside the code.....but maybe it's a good idea....up for debate!

bourdeet: do you mean that the .cfg file should be placed somewhere else?

philippeller: right now we have those under `pisa_examples`

philippeller:
should we just go ahead and get this merged before it is outdated?

bourdeet:
hey! Yes you can merge, unless you really want me to move the config file

PR #647, opened 2020-11-25
nuSQuIDS stage with new nuSQUIDSLayers class

atrettin:
This PR rewrites the nuSQuIDS stage to use the latest developments that I presented a couple of weeks ago. The stage is a bit large, but on the plus side, there are no further sub-modules down the line that you also need to read.

There are multiple new ways to work with it. At the most basic level, one can choose to work in interpolated or in exact mode. 

- **Interpolated mode**: Default mode and how nuSQuIDS is designed to work: Give node positions via a binning with `node_specs` and nodes are placed at the _corners_ of the bins (we don't want to extrapolate and placing nodes at the corners makes that easier). Probabilities are projected out from interpolated interaction picture states between the nodes. This can happen either on another, finer grid or even in events mode controlled by `calc_specs`.
- **Exact mode**: In this mode, a constant density approximation formula is used instead of the numerical integration, and nodes are placed everywhere where we want to evaluate the probability such that there is no interpolation. In this mode, nuSQuIDS behaves just like for example GLoBES with the same trade-off in speed: Much faster for few events, but worse scaling to more events. It also can't support fast oscillation filtering. This mode can be useful for creating pseudo-data without any interpolation and filtering effects in a more reasonable time.

Currently, only standard three flavour and 3+1 oscillations are suppored, but adding decoherence and NSI should be a relatively simple matter. Please let me know if you want this feature and I can help make it happen.

There are also options to filter fast oscillations:
- low-pass filter being applied during numerical integration (hot new feature not yet presented at the call)
- low-pass filter being applied when projecting out probabilities (similar to what has been shown on the call but with some differences)
- production height averaging (as presented on the call). This is not yet implemented but it will be eventually.

The PISA configuration of the new stage is pretty straight-forward, the only specialty is that one needs to define an additional `node_specs` parameter as a MultiDimBinning in addition to `calc_specs`. Nodes are then placed on the _corners_ of the defined bins. Extrapolation is forbidden and it is checked that either all events individually or all bins in `calc_specs` are completely encompassed by the nodes.


atrettin:
Marked as draft because there are a few features still to be implemented.

atrettin:
All of the most important features are now implemented, performance optimized and tested! 🎉 I marked the PR as ready for review.

philippeller:
Hey, I only superficially lookd through the code, but would be happy to merge. Unless anyone else is willing to give it a more thorough review?

PR #651, opened 2021-01-11
Pisa4

philippeller:


philippeller:
Hey! I think we're now getting closer to merging pisa4, yay!
So far I know @atrettin is using it with oscNext and after some fixes works for him. Would be great if other users could test and switch too

atrettin (approved):
Everything seems to be working now, at least for my sterile analysis. There are a couple of fixes still to be made to the nusquids stage and layer calculation, but that can be done after merging this.

philippeller:
Great, thnaks @atrettin !
I was hoping to merge #639 before merging this, but it may just be moving too slow....

PR #654, opened 2021-02-12
Fix Param rescale bug and add log-scaling

atrettin:
This PR fixes a bug where the rescaling gives wrong numbers if the value of a parameter is defined in different units than the range. The fix is to explicitly convert the unit of the ranges to the unit of the value.

I also introduce the option to let the parameter rescale logarithmically, i.e. the _log_ of the parameter's value is rescaled between 0 and 1 during minimization. This can help the minimizer if a parameter can vary across several orders of magnitude. 

Some unit tests are also added to make sure that the true value is always recovered from the rescaled value.


philippeller:
Hi, looks good! Only question: what is the reason behind the log scaling....will this help the minimization process for some params?

atrettin:
@philippeller For my sterile search I'm attempting to fit the mass splitting over several orders of magnitude, and in linear space it can look like the likelihood is very flat for most of the phase space and with a tiny spike at some small number. I'm hoping that the log-scaling makes the gradients more reasonable in that case.

PR #656, opened 2021-02-21
implement mceq systematic with loops

atrettin:
When running the stage, it always produced a numba performance warning saying that the implementation with `np.dot` inside a JIT compiled function is sub-optimal. I implemented the calculation with for-loops instead of `np.dot` and found the computation time for the oscNext high-stats sample drop from 3.7 s to 1.5 s when working in events mode. This is a decent little speed boost for oscillation fits and makes the code easier to understand as well.


philippeller:
works for me. now that we aren't limited to these cuda compatible vectorizers anymore

PR #657, opened 2021-02-25
add option for penalty to metric value in minimizer_callable

marialiubarska:


atrettin:
This looks good, but could you please rebase onto master? It looks like only the last commit is actually new here.

pisa/analysis/analysis.py
```diff
@@ -1190,8 +1190,12 @@ def _minimizer_callable(self, scaled_param_vals, hypo_maker, data_dist,
             fit_history.append(
                 [metric_val] + [v.value.m for v in hypo_maker.params.free]
             )
-            
-        return sign*metric_val
+
+        penalty = 0.
+        if not external_priors_penalty is None:
+            penalty = np.abs(external_priors_penalty(hypo_maker, metric))
```
philippeller, line 1196: why `abs` here? If we need absolute values, i think it should be done inside `external_priors_penalty`

pisa/analysis/analysis.py
```diff
@@ -1190,8 +1190,12 @@ def _minimizer_callable(self, scaled_param_vals, hypo_maker, data_dist,
             fit_history.append(
                 [metric_val] + [v.value.m for v in hypo_maker.params.free]
             )
-            
-        return sign*metric_val
+
+        penalty = 0.
+        if not external_priors_penalty is None:
```
philippeller, line 1195: very superficial comment: usually one writes `if x is not None` instead of `if not x is None` for readability

pisa/analysis/analysis.py
```diff
@@ -1190,8 +1190,12 @@ def _minimizer_callable(self, scaled_param_vals, hypo_maker, data_dist,
...
                 [metric_val] + [v.value.m for v in hypo_maker.params.free]
             )
-            
-        return sign*metric_val
+
+        penalty = 0.
+        if external_priors_penalty is not None:
+            penalty = external_priors_penalty(hypo_maker, metric)
+
+        return sign*metric_val + penalty
```
philippeller, line 1198: Actually, now another question....would `sign` need to be applied to penalty too or not?

marialiubarska: I was assuming that it always should be positive (this is why I had abs there), then it is added to `sign*metric_val` which we always want to minimize.

marialiubarska: However maybe this not the best way: maybe I can add something like `metric_val = metric_val + sign*penalty` inside the if statement and this way avoid adding zero each time if `external_priors_penalty` is not specified

marialiubarska: or we can assume that sign of the penalty is correctly specified inside the `external_priors_penalty` (which has `metric` as an argument, so it can distinguish between llh and chi2)

philippeller: hmm, yeah probably the last point would be the best. Let's do that and maybe you can add a comment into the prior penalty docstring?

marialiubarska: okay, sure!

PR #658, opened 2021-03-03
Nusquids Development and COBYLA support

atrettin:
This is a combined PR where I bring the nuSQuIDS stage up to date with what I am now using for the analysis and also add support for the COBYLA minimizer. The latter didn't turn out to be _that_ great, but it is gradient-free and can work if gradients are terrible. I also made it possible to start minimization at a boundary when using gradient-based optimizers by shifting the starting value just by the step size inward.


philippeller:
Looks good to me, i will merge this

PR #659, opened 2021-03-05
Stacked minimizers

atrettin:
In this PR I introduce a new way to configure an analysis that allows (theretically) for _infinitely_ stackable global search schemes as well as easy extensions for new search strategies!

This is done in a new `BasicAnalysis` class that is only doing minimization and is the new base class for the standard `Analysis`. The core functionality comes from the `fit_recursively` method, which recursively applies global fit strategies that are defined in a dictionary. The trick is that every global fit configuration also passes a dictionary defining how the "local" minimization inside should be done, and this "local" minimization may again contain a global search strategy! 

Currently, the only "global" search strategy implemented is `fit_octants`, which fits two octants (or quadrants, as it were for inflection angles of 90 deg) of a given angle parameter separately. The only "local" fit currently implemented is scipy minimization. But already, this allows for some interesting stacking.

**Examples**

A canonical standard oscillation fit fits octants in `theta23` separately and then runs a scipy minimizer to optimize locally in each octant. The arguments that would produce that result when passed to `fit_recursively` are:
    
```
    method = "fit_octants"
    method_kwargs = {
        "angle": "theta23"
        "inflection_point": 45 * ureg.deg
    }
    local_fit_kwargs = {
        "method": "scipy",
        "method_kwargs": minimizer_settings,
        "local_fit_kwargs": None
    }
```

Let's say we also have a CP violating phase `deltacp24` that we want to fit separately per quadrant split at 90 degrees. We want this done within each quadrant fit for `theta23`, making 4 fits in total. Then we would nest the quadrant fit for `deltacp24` inside the octant fit like so:
    
```
    method = "fit_octants"
    method_kwargs = {
        "angle": "theta23"
        "inflection_point": 45 * ureg.deg
    }
    local_fit_kwargs = {
        "method": "fit_octants",
        "method_kwargs": {
            "angle": "deltacp24",
            "inflection_point": 90 * ureg.deg,
        }
        "local_fit_kwargs": {
            "method": "scipy",
            "method_kwargs": minimizer_settings,
            "local_fit_kwargs": None
        }
    }
```
Finally, let's suppose we want to apply a grid scan over fit starting positions in parameters `theta24` and `deltam41`, and we want to do a standard fit at every grid point with a separate octant fit in `theta23`. This can be done like so:
    
```
    method = "grid_scan"
    method_kwargs = {
        "grid": {
            "theta24": np.geomspace(1, 20, 3) * ureg.deg,
            "deltam41": np.geomspace(0.01, 0.5, 4) * ureg["eV^2"],
        },
        "fix_grid_params": False,  # all parameters free at each grid point
    }
    local_fit_kwargs = {
        "method": "fit_octants",
        "method_kwargs": {
            "angle": "theta23",
            "inflection_point": 45 * ureg.deg,
        }
        "local_fit_kwargs": {
            "method": "scipy",
            "method_kwargs": minimizer_settings,
            "local_fit_kwargs": None
        }
    }
```


atrettin:
Now that several people including myself have configured their analyses using the stacked minimizer features, I think that it's time to get this one into master as well! It looks like there are no more urgently needed features that would have to be implemented.

philippeller:
let me check out your branch and see if the 3y analysis still runs. afterwards i think we can merge unless anyone objects

philippeller:
I get an issue when writing the fit result to a file:

Now the fit result is of type `pisa.analysis.analysis.HypoFitResult`, and i can write it to json, but not reload it. Can you make this class correctly de-serializing?

atrettin:
Ok, I enabled deserialization of the HypoFitResult. It was just a little bit more painful than anticipated, but it works. It is not required with current code in the fridge, because the serialized state is constructed to look exactly like the dictionaries that were produced by the old analysis code, so that fridge plotting stuff should still work without changes.

philippeller:
oops, sorry. Didn't mean to push to your branch directly!
Had to handle a small exception. Otherwise I'm happy with the PR now

atrettin:
Great, thank you for looking into it! Then from my side we can pull it.

PR #663, opened 2021-04-26
Expanding Hyper-Hypersurfaces to N dimensions

atrettin:
This is a major rework of hypersurfaces to facilitate large-scale interpolation in N dimensions, which became necessary because I found that I need to interpolate at least in 3D for the sterile neutrino analysis. The previous code could only support 2D at most, and the grid points were all loose files being put into a special folder structure which would have become a nasty thing going to higher dimensions. The goal of this rework is also to make the logistics around fitting hypersurfaces on a grid as simple as possible for the user and to store all of the results in one large JSON file that can be read in one go.

There are basically three changes:
* hypersurface code split up into three files
* `HypersurfaceInterpolator` supports ND interpolation via scipy's `RegularGridInterpolator`
* rework of the entire scaffolding around fitting a grid of hypersurfaces on a cluster and merging all the results to a `HypersurfaceInterpolator`. 

The code for standard non-interpolating hypersurfaces remains unchanged.

The new procedure for fitting hypersurfaces on a grid is demonstrated in the [fitting script for the sterile analysis](https://github.com/icecube/wg-oscillations-fridge/blob/master/analysis/oscnext_sterile/analysis/fit_oscnext_hypersurfaces.py). It only requires running three functions:
1. `prepare_interpolated_fit` takes in all of the information about nominal and systematic sets and a fit directory. It dumps all of the necessary information to run each fit into JSON files in the given directory.
2. `run_interpolated_fit(fit_dir, job_idx)` is run by each job on the cluster. All that is needed is the fit directory and the job index, and thanks to the prepared JSON files all of the other information to run the fit can be loaded automatically.
3. `assemble_interpolated_fits(fit_directory, output_file)`: When all jobs are complete, this function automatically scans the `fit_directory` for completed fits and assembles them all into one large JSON named `output_file`. This is then the file that is passed to the PISA stage as `fit_results_file`.

The plotting function `plot_fits_in_bin` of the `HypersurfaceInterpolator` is also improved and can now show 1D and 2D slices of interpolated fits as long as a value for the remaining dimensions is chosen. The function is demonstrated in a [notebook in the fridge](https://github.com/icecube/wg-oscillations-fridge/blob/master/analysis/oscnext_sterile/notebooks/plot%20hypersurface%20interpolation.ipynb).


atrettin:
Here is an example of a 2D slice taken from a 3D interpolated hypersurface from one bin in the sterile analysis. The interpolation runs over the parameters `deltam31`, `deltam41` and `theta24`, and `deltam31` was set to 2.5e-3 eV^2 for this slice. 

hs_slice_deltam31_2 5e-3eV2_bin_8-8-2

marialiubarska (commented):
Hi @atrettin ! Sorry for the delay, I finished reviewing the code.
Overall everything looks good to me, I just have a few minor questions:

1. in `make_slices` in `pisa/utils/hypersurface/hyper_interpolator.py` does [this loop](https://github.com/icecube/pisa/blob/a8ebb21f0ad34715d696d34760bb88cd718a3d56/pisa/utils/hypersurface/hyper_interpolator.py#L280) assume that `xi` is ordered in the same way as `self.interpolation_param_names`? This is the case when it is [used](https://github.com/icecube/pisa/blob/a8ebb21f0ad34715d696d34760bb88cd718a3d56/pisa/utils/hypersurface/hyper_interpolator.py#L347) in `plot_fits_in_bin`, but if, for some reason, user tried to call `make_slices` directly with the different order of parameters, it looks like parameter values can get mixed. Would it make sense to check not only the length of `xi`, but also the order of parameters?
2. in `plot_fits_in_bin` there is an [assertion](https://github.com/icecube/pisa/blob/a8ebb21f0ad34715d696d34760bb88cd718a3d56/pisa/utils/hypersurface/hyper_interpolator.py#L310) that `plot_dim <= 2`. Do you think it would make sense to explicitly check if `plot_dim in [1,2]` instead? This is not a very probable case, but technically someone can pass more parameters with `**param_kw` then `self.ndim` and then `plot_dim` will be zero or negative. This is not a huge deal, because then the code will fail [here(?)](https://github.com/icecube/pisa/blob/a8ebb21f0ad34715d696d34760bb88cd718a3d56/pisa/utils/hypersurface/hyper_interpolator.py#L329), but it would be good to have a proper error for this case.
3. this is a very damp question, but when I'm looking at the [creation](https://github.com/icecube/pisa/blob/a8ebb21f0ad34715d696d34760bb88cd718a3d56/pisa/utils/hypersurface/hyper_interpolator.py#L798) of `OrderedDict` for `metadata` in prepare_interpolated_fit`, I can't figure out how it would work. Is it possible pass entries as parameters like this? 
4. I don't have a good suggestion for the [while True loop](https://github.com/icecube/pisa/blob/a8ebb21f0ad34715d696d34760bb88cd718a3d56/pisa/utils/hypersurface/hyper_interpolator.py#L561) in `hyper_interpolator.py`, but I think alternatively one can use os.listdir to get all the gridpoint files in the directory if that's what they prefer
5. I also have a comment on the `hypersurfaces.py` stage, which is not directly connected to this PR, but I thought it would make sense to mention it here. There is an [assertion](https://github.com/icecube/pisa/blob/a8ebb21f0ad34715d696d34760bb88cd718a3d56/pisa/stages/discr_sys/hypersurfaces.py#L100) that checks if `self.calc_mode` is an instance of `MultiDimBinning`. If this is not the case, and regular (not interpolated) hypersurfaces are used, the code will fail earlier on the similar [assertion](https://github.com/icecube/pisa/blob/a8ebb21f0ad34715d696d34760bb88cd718a3d56/pisa/utils/hypersurface/hypersurface.py#L1640) in `load_hypersurfaces`. Would it make sense to move the assertion in the stage to the point before the hypersurfaces get loaded?
6. Finally, a very nitpicky question: in the `hypersurfaces`  stage definition parameter `interpolated` is missing from the docstring, is this intentional?

I tried my best to understand the code, but it is totally possible I got something wrong, so please feel free to correct me!
I'm also tagging @philippeller in case he would also like to comment on some of these questions.

philippeller:
I have a more high-level question. With the changes to the hyper surface stage, are we still able to run this with existing hypersurfaces? I.e. does it preserve backwards compatibility?

atrettin:
Thank you for the in-depth review, @marialiubarska ! I know it's a lot of stuff... I still didn't have much time to work through it, but there are a few things I'm going to patch.

Concerning the questions:
1. The problem here is that `xi` are already floats, so there is no way to tell the order at this point. Maybe I should make the method private since there aren't so many reasons I can think of that one would want to call it from the outside....
2. Good point, it would be better to make the check more explicit like that!
3. Yes, that is a valid syntax to make an `OrderedDict`. It ensures that the keys appear in the order that they are entered, though I think at that point it wouldn't matter. 
4. One problem that I found myself is that, if you fit again with a different grid in the same directory, this function would get confused. It would be better to explicitly calculate the expected number of grid points from the grid specification.
5. Good point, the assertion should be made as early as possible.
6. I forgot that one... Good catch! :) 

@philippeller For old, non-interpolated hypersurfaces the answer is yes. For interpolated ones, no, because the storage structure has fundamentally changed. Basically, the entire management of how to arrange the files used to be in the fridge and is now in PISA. But it's not a big deal, in my mind, re-fitting is pretty quick and we don't have legacy data to take care of.

atrettin:
@marialiubarska I made some changes to address the concerns you raised. Thanks again for checking!

marialiubarska (approved):
Hi @atrettin , thank you very much! I looked at the changes and I don't have any more concerns.

@philippeller I'm approving this PR

PR #668, opened 2021-07-13
Fixes to make unit tests run without failure

atrettin:
This PR adds `__deepcopy__` to `MapSet` that manually creates a new `MapSet` objects with deep copies of all maps therein. The unit test checks that the copy and the original are identical and that the copy and the original really are decoupled.

This avoids the crash of the unit test when attempting to deep-copy a `MapSet`.


atrettin:
Converted to draft to add more fixes.

philippeller:
first fix look good to go
I'll wait for what else is to come....

atrettin:
Okay that was already it it should now work. I don't quite understand why it was necessary myself but import `*` did not find the one function?

atrettin:
Okaaaay it looks like our unit tests were completely meaningless the entire time because even an `AssertionError` would not let the script exit with a failure code that GitHub recognizes. 😅

atrettin:
The exit code has to be some thing [other than zero](https://docs.github.com/en/actions/creating-actions/setting-exit-codes-for-actions) for the action to be considered failed. TIL.

atrettin:
Tadaa! We now have meaningful unit tests and they don't fail! 🎉 There are still tons of warnings for example that the "target" keyword is deprecated in numba but that's not critical right now.

philippeller:
Thanks @atrettin for looking into this and fixing multiple things on the way! I browsed through the changes, and they look good to me, so i will merge this

PR #671, opened 2021-08-04
Analysis fix

mvprado0:
Fixed non-list metrics issue and included functions in stats.py to better condense some lines of code.

### DIFF
--- a/pisa/analysis/analysis.py
+++ b/pisa/analysis/analysis.py
@@ -823,7 +823,7 @@ class BasicAnalysis(object):
 
         # Take the one with the best fit
         got_better = it_got_better(new_fit_info.metric_val, best_fit_info.metric_val, metric)
-        
+
         # TODO: Pass alternative fits up the chain
         if got_better:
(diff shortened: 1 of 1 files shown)


philippeller:
Hi, could you comment how the bound guard is now needed after @atrettin bumped the scipy version which fixed the boundary issue?

mvprado0:
Hmm that's odd. I actually didn't do anything with the bound guards. I only implemented a couple of functions found in stats.py that are completely separate and fixed the issue that arises if the user inputs a metric that is not a list. 
I see the part about bound guards that you're talking about but I'm confused why that changed. Maybe I somehow forked an older version of analysis.py? I didn't think I did...

mvprado0:
I just rechecked and I only changed the things until line 1184. Anything after that with the bound guard I didn't touch. So I'm confused

mvprado0:
Ok, all fixed. I reincorporated my new changes to Alex's guard bounds commit.

atrettin (approved):
This looks good to me now!

PR #673, opened 2021-08-20
Implement fast histogramming and lookup

atrettin:
This PR implements fast histogramming with the `fast_histogram` package and fast lookup with dedicated numba functions for regular binnings. 

When binnings are fully regular and linear, the `fast_histogram` library is automatically used when Container contents are translated from arrays to histograms, and the fast lookup functions are used when Container contents are translated from binnings to arrays. 

Because the logarithm is a very costly operation (especially in double precision), regularizing binnings into log-space would be very slow and nearly negate the advantage from using fast histograms. To circumvent this problem, a mechanism is implemented in the Container class to cache the logarithm of array representations. There are now two array representations: "events" and "log_events". When the "log_events" representation is requested, the log is calculated event-wise. As with any other representation, the result is cached such that the log need not be calculated again unless the contents are changed via item assignment.

With these two changes, histogram and lookup operations for regular (linear or logarithmic) binnings are accelerated by a factor between 5-10, at least in artificial benchmarks with normally distributed data, without any need to change Pipeline configurations or stages. 

Finally, the `hist` stage automatically detects irregular axes in the output binning and pre-computes indices for these axes (typically the PID axis) such that a regular binning in the index can be used. This is all done automatically, such that every analysis, even one using an irregular output binning, can take advantage of fast histogramming.

Note
-------
Currently, not all features needed have been merged into the `fast_histogram` package for a release. For now, an installation from the source of [my fork's "xmax_fix" branch](https://github.com/atrettin/fast-histogram/tree/xmax_fix) is required. I'm pushing for an official release, but until then, I will keep this a draft.

### DIFF
--- a/pisa/core/binning.py
+++ b/pisa/core/binning.py
@@ -1841,6 +1841,21 @@ class MultiDimBinning(object):
         """Return a list of the contained dimensions' bin_edges that is
         compatible with the numpy.histogramdd `hist` argument."""
         return [d.bin_edges for d in self]
+    
+    @property
+    def is_irregular(self):
+        """Returns `True` if any of the 1D binnings is irregular."""
+        return np.any([d.is_irregular for d in self])
+    
+    @property
+    def is_lin(self):
+        """Returns `True` iff all dimensions are linear."""
+        return np.all([d.is_lin for d in self])
+    
+    @property
+    def is_log(self):
+        """Returns `True` iff all dimensions are log."""
+        return np.all([d.is_log for d in self])
 
     @property
     def domains(self):
--- a/pisa/core/container.py
+++ b/pisa/core/container.py
@@ -20,6 +20,7 @@ from pisa.core.map import Map, MapSet
 from pisa.core.translation import histogram, lookup, resample
 from pisa.utils.comparisons import ALLCLOSE_KW
 from pisa.utils.log import logging
+from pisa.utils.profiler import line_profile, profile
 
 
 class ContainerSet(object):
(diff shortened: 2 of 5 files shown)


atrettin:
Now that all unit tests pass, this could actually be pulled! 🎉 We can still change the `setup.py` whenever the actual release of `fast_histogram` is made, but for now this seems to be ok as it is.

philippeller:
Great! Does the fast histo package now compile automatically during installation?

Also, just out of interest, how much of a speed-up do you get with this for you analysis?

atrettin:
> Great! Does the fast histo package now compile automatically during installation?

Yes! Everything should just happen magically! 🌈✨🦄 It works in the GitHub workflow, so I hope it will work anywhere. 

> Also, just out of interest, how much of a speed-up do you get with this for you analysis?

Sadly, the speedup in the real world is more modest than what artificial tests with normally distributed data suggested. When I benchmark the `Container` class on its own, I get a speedup of 10, but the real speedup in the `apply_function` of the `hist` stage is a factor of 3.5 for an absolute time saving of 500 ms, the speedup of the `apply_function` in the nusquids stage is 5 for a saving of 800 ms.

An "idle" output from my PISA pipeline (i.e. without changing any parameter, just going through all the `apply` functions) is sped up by a factor of 2 from 900 ms to 450 ms (the timings varied unexpectedly wildly, some of these numbers may seem inconsistent). An output after changing one parameter in nusquids now takes 4.7 s where it used to take 6 s. A modest speedup, but not a game-changing one.

PR #674, opened 2021-08-26
Set number of threads globally with PISA_NUM_THREADS variable

atrettin:
This PR allows one to set the number of threads to be used for multithreading with one central environment variable `PISA_NUM_THREADS`. This variable is used by `numba` as well as any "custom" multithreading implementation (of which there is currently only one, for fast histograms). 

Multithreading can be made much more simple than we used to: All one needs to do is to implement things with simple for-loops and then use numba's `prange` instead of `range` in for-loops that can be naively parallelized. This is done for the `mceq_barr_red` stage as well as the fast lookup procedures. I would recommend going this route as well for `prob3`, but I don't think that I'm going to do this because I'm not using it myself. I had to remove the `target` keyword in some places because it no longer works.

nuSQuIDS already support multithreading for the most expensive operation of propagating state densities through matter potentials, and here, too, the `PISA_NUM_THREADS` variable is used.

Multithreading for fast histograms is possible because the Global Interpreter Lock is released during the main loop, allowing Python to continue during the calculation. The input sample is simply split up in chunks and the histogram is calculated for every chunk, and the outputs are added together. The speedup is not entirely a factor of N for N cores, but there is benefit in it.

### DIFF
--- a/pisa/__init__.py
+++ b/pisa/__init__.py
@@ -216,14 +216,15 @@ del cpu_targets, gpu_targets, parallel_targets
 PISA_NUM_THREADS = 1
 """Global limit for number of threads"""
 
-# ignore if target is not multicore
-if TARGET == 'parallel':
-    if 'PISA_NUM_THREADS' in os.environ:
-        PISA_NUM_THREADS = int(os.environ['PISA_NUM_THREADS'])
-        assert PISA_NUM_THREADS >= 1
-    else:
-        PISA_NUM_THREADS = numba.config.NUMBA_NUM_THREADS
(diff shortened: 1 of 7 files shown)


philippeller:
Would be good to throw a warning in `__init__.py` for when users set `PISA_NUM_THREADS` > 1, but `PISA_TARGET` = cpu

philippeller:
otherwise looks to be working correctly

atrettin:
> Would be good to throw a warning in `__init__.py` for when users set `PISA_NUM_THREADS` > 1, but `PISA_TARGET` = cpu

Okay, done!

PR #675, opened 2021-08-30
Fixes for test notebook 

JanWeldert:
Bring unit back to container and correctly check if binning needs to be log


philippeller:
@atrettin  Are you ok with the above changes?

atrettin:
> @atrettin Are you ok with the above changes?

Yes.

PR #676, opened 2021-08-31
Revive detectors class

JanWeldert:
I updated the detectors class and added a test notebook. A few changes to analysis.py were also needed.

And I clearly do something wrong merging master. The only interesting commit is the second to last one.


philippeller:
Changes looking good and uncontroversial to me, going to merge

PR #677, opened 2021-09-17
Modified mceq barr stage to use for-loops.

LeanderFischer:
I modified the full barr stage to use for-loops, which gives a small speedup. When running the full oscnext high stats neutrino pipeline it improves from ~12.3 s to ~8.2 s. The (very simple) notebook for that is [here](https://github.com/icecube/wg-oscillations-fridge/tree/master/analysis/sandbox/lfischer/test_barr_stage).
There are some additional modifications similar to the changes in the reduced barr stage.

* updated docstring
* moved antipion_production function into stage
* linked containers for speedup
* added check for completeness of spline file
* converted barr parameters application into for loops

### DIFF
--- a/pisa/stages/flux/mceq_barr.py
+++ b/pisa/stages/flux/mceq_barr.py
@@ -20,21 +20,16 @@ from numba import guvectorize
 from pisa import FTYPE, TARGET
 from pisa.core.stage import Stage
 from pisa.utils.log import logging
-from pisa.utils.profiler import profile
+from pisa.utils.profiler import profile, line_profile
 from pisa.utils.numba_tools import WHERE, myjit
 from pisa.utils.resources import find_resource
 
 
-def antipion_production(barr_var, pion_ratio):
-    """
-    Combine pi+ param and pi+/pi- ratio to get pi- param
-    """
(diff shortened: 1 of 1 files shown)


pisa/stages/flux/mceq_barr.py
```diff
@@ -241,6 +247,18 @@ def setup_function(self):
         # Encoding is to support pickle files created with python v2
         self.spline_tables_dict = pickle.load(BZ2File(spline_file), encoding="latin1")
 
+        # Ensure that the user is not loading an incompatible spline
```
philippeller, line 250: I think to remember that this linking had to be disabled in the past for this stage due to some observed differences, which were unresolved at the time. Maybe this has now gone away with pisa4 though? Would be good to test

LeanderFischer: The linking in line 184-193? What were the observed differences? I can take a look. I added these lines comparing to [the reduced barr stage](https://github.com/icecube/pisa/blob/master/pisa/stages/flux/mceq_barr_red.py), which also has the linking.

philippeller: Hmm, maybe my memory is tricking me. Will merge this then

PR #678, opened 2021-09-24
Multithreading fix

atrettin:
Because the nuSQuIDS stage didn't check whether the `TARGET` was `parallel`, it would use multicore despite the target being CPU. We now check whether the target is `parallel` and _in addition_ make sure that the global variable `PISA_NUM_THREADS` is always 1 if `TARGET` is `cpu`.

We also only give out the warning if the user tried to set `PISA_NUM_THREADS` in their environment explicitly larger than 1, but the `TARGET` is `cpu`.


philippeller:
looks fine

Issue #649, opened 2020-12-08
Iminuit 2


philippeller:
Breaking changes in v2,need to update hypersurface fitting

atrettin:
This issue is now resolved.

PR #681, opened 2021-10-20
Support iminuit 2.x for hypersurfaces and local fits

atrettin:
This PR fixes issue #649 and adds iminuit as yet another option to the toolkit of the Analysis class. It can be dropped in place of local scipy optimization by setting `local_fit_kwargs` to a dictionary with the following content:

```
minuit_settings = {
    "errors": 0.1,  # initial Migrad step size, not very important
    "precision": 1e-14,  # default: double precision
    "tol": 0.1,  # default: 0.1
    "run_simplex": True  # run simplex before polishing the result with Migrad (default: False)
    "run_migrad": True  # run migrad (default: True). If `False`, runs only Simplex and exits
}

local_fit_kwargs = {
    "method": "iminuit",
    "method_kwargs": minuit_settings,
    "local_fit_kwargs": None  # minuit is a local optimization and therefore ends the recursion
}
```

The `errordef` option is automatically set depending on the type of loss function being optimized.
When both Simplex and Migrad are run, Simplex stops at a relatively lax tolerance and the result is polished with Migrad.

### DIFF
--- a/pisa/analysis/analysis.py
+++ b/pisa/analysis/analysis.py
@@ -18,6 +18,7 @@ import numpy as np
 import scipy.optimize as optimize
 # this is needed for the take_step option in basinhopping
 from scipy._lib._util import check_random_state
+from iminuit import Minuit
 
 import pisa
 from pisa import EPSILON, FTYPE, ureg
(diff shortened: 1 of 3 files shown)


mvprado0 (commented):
Everything looks good to me!

PR #685, opened 2021-11-11
Add staged fit

atrettin:
Add a module to perform a staged fit, where a later fit starts from the best fit point of an earlier fit. 

The use case for this is primarily to polish the result of a global search. For example, I can run NLOPT's CRS2 global search algorithm for up to 30h and then polish the best fit point using L-BFGS-B with the following kwargs to `fit_recursively()`:

```
method="staged",
method_kwargs=None,
# List of subsidiary fits must at least have length 2.
# Fits are run in order of the list.
local_fit_kwargs=[
    fit_nlopt_crs2,
    local_fit_scipy,
]
```
where
```
fit_nlopt_crs2 = collections.OrderedDict(
    method="nlopt",
    method_kwargs={
        "algorithm": "NLOPT_GN_CRS2_LM",
        "ftol_rel": 1e-7,
        "ftol_abs": 1e-5,
        "maxtime": 30 * 3600,
    },
    local_fit_kwargs=None,
)
```
and 
```
scipy_settings_l_bfgs_b = {
  "method": {
    "value": "L-BFGS-B",
    "desc": "The string to pass to scipy.optimize.minimize so it knows what to use"
  },
  "options":{
    "value": {
      "disp"   : 0,
      "ftol"   : 1.0e-7,
      "eps"    : 1.0e-7,
      "gtol"   : 1.0e-8,
      "maxiter": 10000
    },
    "desc": {
      "disp"   : "Set to True to print convergence messages",
      "ftol"   : "Precision goal for the value of f in the stopping criterion",
      "eps"    : "Step size used for numerical approximation of the jacobian.",
      "maxiter": "Maximum number of iteration"
    }
  },
}
local_fit_scipy = {
    "method": "scipy",
    "method_kwargs": scipy_settings_l_bfgs_b,
    "local_fit_kwargs": None
}
```

One issue that I hope won't cause problems is the fact that I set the nominal parameters of the hypo_maker to the best fit values of the previous fit. This is done to make sure that a subsidiary fit doesn't destroy all progress by calling `reset_free()`, which could easily happen by mistake since it is the default for some modules. We should test whether or not this creates problems with different param selections similar to what @ts4051 has found elsewhere.

### DIFF
--- a/pisa/analysis/analysis.py
+++ b/pisa/analysis/analysis.py
@@ -1357,7 +1357,50 @@ class BasicAnalysis(object):
         original_param.value = best_fit_result.params[original_param.name].value
         hypo_maker.update_params(original_param)
         return best_fit_result
-
+    
+    def _fit_staged(self, data_dist, hypo_maker, metric,
+                    external_priors_penalty, method_kwargs, local_fit_kwargs):
+        """Run a staged fit of one or more sub-fits where later fits start where the
+        earlier fits finished.
+        
+        The subsidiary fits are passed as a list of dicts to `local_fit_kwargs` and 
+        are worked on in order of the list. Internally, the `nominal_values` of the 
+        parameters are set to the best fit values of the previous fit, such that
+        calls to `reset_free` do not destroy the progress of previous stages.
+        """
+        assert local_fit_kwargs is not None
+        assert isinstance(local_fit_kwargs, list) and len(local_fit_kwargs) > 1
+        
+        logging.info("Starting staged fit...")
+        best_fit_params = None
+        best_fit_info = None
+        # storing original nominal values
+        original_nominal_values = dict(
+            [(p.name, p.nominal_value) for p in hypo_maker.params.free]
+        )
+        for i, fit_kwargs in enumerate(local_fit_kwargs):
+            logging.info(f"Beginning fit {i+1} / {len(local_fit_kwargs)}")
+            if best_fit_params is not None:
+                hypo_maker.update_params(best_fit_params)
+            best_fit_info = self.fit_recursively(
+                data_dist, hypo_maker, metric, external_priors_penalty,
+                fit_kwargs["method"], fit_kwargs["method_kwargs"],
+                fit_kwargs["local_fit_kwargs"]
+            )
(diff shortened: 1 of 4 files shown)


atrettin:
Rebased onto master. I discovered a bug after including the changes from the latest merge: When one of the fits in the staged fit is an octant-fit (`fit_octants`), then it crashes at L1397 in analysis.py because the range of the angle parameter was not reset.

atrettin:
- Re-based onto the latest changes to correctly handle memory pointers from param selectors. 
- The analysis functions now always update the values inside the parameters without changing which parameters are being pointed to. 
- Added direly needed unit test for the recursive fit strategy. It runs one ridiculously complicated fit that touches almost all fit modules and asserts that this does not result in inconsistencies in the stored parameters.

atrettin:
Finally, unit tests are working! This was a huge technical debt on my side, but now we can have at least a little bit of confidence in this thing. :)

pisa/analysis/analysis.py
```diff
@@ -320,6 +320,40 @@ def get_separate_octant_params(hypo_maker, angle_name, inflection_point) :
 
     return angle_orig, angle_case1, angle_case2
 
+def update_param_values(
```
ts4051, line 323: It would be great if this function could also accept a `Pipeline` instance as well as a `DistributionMaker`, to make it usable for simpler cases where only a pipeline is used

ts4051 (commented):
I have reviewed the code now. Both the concept and implementation of the new functionality looks good to me, and the commenting is also clear. The new testing is also great to have. I have no major changes to suggest.

pisa/analysis/analysis.py
atrettin: I added support for passing single Pipeline objects and unit test that it works correctly.

atrettin:
Tiny addition: Exposing the tolerance setting for octant fits. In some cases where the likelihood at the inflection point is very flat and the truth is far away, I might want to offset a little more into the octant (or quadrant in this case, the pathological angle is the sterile CP phase with inflection point at 90 degrees).

PR #687, opened 2021-11-26
Correct handling of selectable params

philippeller:
Before, only the current representation of the selectable params would get overwritten in the distributionmaker instantiation, such that params of the same name would point to the same object.
Furthermore, the update function in the pipeline would call update on the params instead of the selector.
Also, the update in param selector would ignore the values in the underlying dict


atrettin (approved):
This works! I can now switch between param selections even if some (but not all) parameters have duplicates in several pipelines.

PR #688, opened 2021-12-08
Fix small parameter scaling issues

atrettin:
In some special pathological cases, rounding errors can cause a rescaled value to end up outside the bounds of a parameter. In this case, we want to clip the value back to the edge of the range.

Another small issue is that some functions rely on the fact that `range[0]` is the smaller value and `range[1]` is the larger value. Because this is a thing that many users might rely on elsewhere, we can avoid hard-to-catch numerical errors by enforcing that order in the setter method of the range.


philippeller:
Thanks for this PR, it looks good to me - will merge!

PR #690, opened 2022-02-02
Make it possible to weight events

JanWeldert:
Of course events are already weighted, but what I mean is to use a weight per event representing reconstruction uncertainties. In order to do that we have to modify the metric (called weighted_chi2 now) and tell the hist stage to compute the bin uncertainties correctly. In case of no uncertainty weights the weighted_chi2 is identical to the mod_chi2.


JanWeldert:
btw, I have no idea why so many of my old commits are listed here. Only the last is important

atrettin:
I'm not sure I understand how the weight represents reconstruction uncertainties... could you please explain what is happening and why? I'm confused why you say that the uncertainty takes reconstruction uncertainty into account, yet when I read the code of the loss function it looks like it is only about the MC uncertainty... which we already have in the `mod_chi2` loss. How is `weighted_chi2` different from `mod_chi2`?

JanWeldert:
You can now add per event uncertainty weights to a container. The difference to the other MC weights is that they don't represent a number of events, but really a weight that would also be included to data in the same way. As a consequence you can't used Poisson errors for the analysis bin contents anymore

atrettin:
Hm, ok. Is the idea, then, that events with a higher uncertainty get less weight or something to the effect?

JanWeldert:
Correct, the uncertainty weight could be something like exp(-|uncertainty|) or so

philippeller (commented):
Hi, sorry for this sitting around stale for such a long time! The changes look in order to me without those impacting anything else. So from my side this can be merged

philippeller (approved):

pisa/analysis/analysis.py
```diff
@@ -2456,6 +2477,7 @@ def _minimizer_callable(self, scaled_param_vals, hypo_maker, data_dist,
         #
         try:
             if hypo_maker.__class__.__name__ == "Detectors":
+                hypo_maker.update_params(hypo_maker.params)
```
atrettin, line 2480: I don't understand this line... why update the parameters with the parameters that it already has? Be sure that this is what you want: The `update_params` method replaces the references inside the `hypo_maker`, which can lead to unintended consequences! If you intend to change the *values*, please use the `update_param_values` function defined in this module.

pisa/analysis/analysis.py
```diff
@@ -619,15 +619,25 @@ def get_detailed_metric_info(data_dist, hypo_asimov_dist, params, metric,general
...
-                    expected_values=hypo_asimov_dist, metric=m
-                )
-                metric_hists = data_dist.metric_per_map(
-                    expected_values=hypo_asimov_dist, metric='binned_'+m
-                )
+                if m == 'weighted_chi2':
+                    actual_values = data_dist.hist['total']
+                    expected_values = hypo_asimov_dist.hist['total']
+                    d = {'output_binning': hypo_maker.pipelines[0].output_binning,
+                         'output_key': 'bin_unc2'}
```
atrettin, line 629: Just so i understand: `bin_unc2` is something that is provided at some point earlier in the pipeline for this to work, right?

atrettin (commented):
Looks good to me, but have a look at the `update_params` call and make sure that it doesn't lead to a loss of object references.

pisa/analysis/analysis.py
JanWeldert: Correct, that is calculated in `pisa/stages/utils/hist.py` based on per event uncertainties provided to the pipeline

pisa/analysis/analysis.py
atrettin: @JanWeldert Could you please come back to this? It is my last question, once we have sorted this line out we are good to go.

atrettin: Ok, I see now. I have never worked with multiple detectors before, but at least from a cursory reading it checks out.

atrettin:
Maaaaybe it would be nice to add at least a small multi-detector example to the unit test of the Analysis class, since that case hasn't been tested before at all. Probably you could just copy&paste from that testing notebook to do it? But if that would be very complicated we don't have to, I would also be ok with merging this.

JanWeldert:
I can definitely try :)

JanWeldert:
Does this look good?

atrettin:
> 

Thank you for getting to this so quickly! Is the test function found and run automatically in our testing pipeline on GitHub? I'm not sure about it, Pytest would probably find it due to the name, but we don't use Pytest for test discovery (which is unfortunate!). I think you need to move the test function into the script that defines the module (pisa/core/detectors.py) so that the test will be found and run automatically.

atrettin (approved):
With unit tests passing, I approve of these changes. If there are unrelated issues with the Detectors class, we will deal with them separately. LGTM!

PR #691, opened 2022-02-17
New stages for high-E analyses, some changes to existing ones

BenSmithers:
A lot of these additions are a work-in-progress, but I didn't want my fork getting too out-of-sync with the master branch! Also want to be sure my changes so far don't cause any conflicts. 

## Changes

`stages/osc/nusquids.py`

 -> Including taus now, added toggle to turn taus on or off. Defaults to off

`pisa/scripts/create_barr_sys_tables_mceq.py`

-> now also calculates the gradients including tau contributions 

`stages/utils/hist.py`

-> when getting weights, now checks if there is an `astro_weights` container. Sums `weights` and `astro_weights` containers if so

`utils/flux_weights.py`

-> a bool can be passed to the `load_2d_honda_table` function to switch behavior between the standard honda table files and the hillas-gaisser h3a flux tables (see additions) 

-> Also modified the `load_2d_table` function to allow for flux files with "hillas" in the name instead of only "honda" or "bartol". These fluxes are loaded in and splined similarly to how the honda fluxes are. 

## Additions

`stages/flux/hillasg.py`

->Almost identical to flux stage `honda_ip`. This explicitly requries three flavors in the flux. Added as separate stage since making the honda_ip one backwards-compatible seemed ugly

`stages/data/meows_loader.py`

-> stage to load in an hdf5 file with all the MEOWS MC

`stages/flux/airs.py`

-> Implements new parameter from atmospheric uncertainty splines

`stages/flux/astrophysical.py`

-> Incorporates a separately weightable astrophysical neutrino weight. Adds parameters for the normalization and spectral index uncertainties

`stages/data/licloader_weighter.py`

Still a work in progress, this is to load in LeptonInjector MC

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -102,7 +102,7 @@ Required Python modules that are installed automatically when you use the `pip` 
 
 Optional dependencies. Some of these must be installed manually prior to installing PISA, and some will be installed automatically by pip, and this seems to vary from system to system. Therefore you can first try to run the installation, and just install whatever pip says it needed, or just use apt, pip, and/or conda to install the below before running the PISA installation.
 
-* [LeptonInjector](https://github.com/icecube/leptoninjector) Required for the `data.licloader_weighter` service. 
+* [LeptonWeighter](https://github.com/icecube/leptonweighter) Required for the `data.licloader_weighter` service. 
 * [MCEq](http://github.com/afedynitch/MCEq) Required for `flux.mceq` service.
 * [nuSQuiDS](https://github.com/arguelles/nuSQuIDS) Required for `osc.nusquids` service.
 * [pandas](https://pandas.pydata.org/) Required for datarelease (csv) stages.
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -102,11 +102,13 @@ Required Python modules that are installed automatically when you use the `pip` 
 
 Optional dependencies. Some of these must be installed manually prior to installing PISA, and some will be installed automatically by pip, and this seems to vary from system to system. Therefore you can first try to run the installation, and just install whatever pip says it needed, or just use apt, pip, and/or conda to install the below before running the PISA installation.
 
+* [LeptonInjector](https://github.com/icecube/leptoninjector) Required for the `data.licloader_weighter` service. 
 * [MCEq](http://github.com/afedynitch/MCEq) Required for `flux.mceq` service.
 * [nuSQuiDS](https://github.com/arguelles/nuSQuIDS) Required for `osc.nusquids` service.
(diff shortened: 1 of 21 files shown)


BenSmithers:
It looks like it failed while importing stages relying on photospline, LeptonWeighter, and/or nuSQuIDS

philippeller:
Great, thanks for the PR, i will take a look shortly.
About the errors, I'd just make your code exit gracefully if leptonweighter et. al cannot be imported. (try import, except warn and terminate)

BenSmithers:
Is that preferable to adding LeptonWeighter and Photospline to the list of optional modules in the unit tests? If so I'll go ahead and push a commit to this PR so the relevant scripts terminate with a `sys.exit()` if they fail to import a module.

philippeller:
Ah, if those can be declared to unit test to not make it fail, it will be even better. Do you want to give this a try?

Also, the optional modules should be mentioned here: https://github.com/icecube/pisa/blob/master/INSTALL.md

BenSmithers:
There we go. Updated the INSTALL file and added photospline, LeptonWeighter, and the alternate spelling of `nuSQuIDS`' python package to the ok-to-ignore list. Sorry for the test-fail spam after all the silly mistakes in my commits!

philippeller:
Great! The new stages and code changes look good to me, so from my side we should be able to merge.
But let's give @atrettin a chance to comment too

philippeller (approved):

atrettin (approved):
The changes look good to me (I mostly looked at nuSQuIDS and histogramming). As the changes are implemented, they shouldn't interfere with anything else if there are no taus, so that seems all safe. A small style nitpick I have is that there are some files that still have `from __future__ import` stuff in them from back when we were still supporting Python 2, which we no longer do. I would also prefer if we could use `black` to make styles more uniform (there are often inconsistencies in spacing, use of parentheses, etc.). But that is all optional and I approve of this PR as it is now.

philippeller:
Great! If @BenSmithers could make these small changes suggested by @atrettin (remove future imports, run through black formatter), we will merge this PR

philippeller:
Sorry we just merged #689 now resulting in a conflict. So one more todo before we merge this one

BenSmithers:
I removed the `from __future__` imports from and applied the black formatter on the files I worked with. Doing so on the rest of the repository (and without messing with the git-blame) seems like it would be quite substantial, and I think it would merit a PR on its own.

philippeller:
cool! Yeah if you are eager to run black over the whole repo, please make a separate PR for it! :)
Going to merge this one now

PR #693, opened 2022-02-25
Sqlite dataloader

RasmusOrsoe:
Hi!

This pull request contains a sqlite data loader that is compatible with the sqlite reconstruction pipeline from GraphNeT.

The loader is tested using the 3year oscillation example and does indeed work.

Rasmus


pisa/stages/data/sqlite_loader.py
```diff
@@ -0,0 +1,152 @@
...
+    database : path to sqlite database
+    **kwargs
+        Passed to Stage
+    """
+    def __init__(
+        self,
+        database,
+        output_names,
+        add_these =  ['energy', 'zenith'],
+        is_retro = False,
```
philippeller, line 28: could we here simply specify `_postfix` to make the stage able to load any reco.....for example also FLERCNN etc from the sqlite database?

pisa/stages/data/sqlite_loader.py
```diff
@@ -0,0 +1,152 @@
...
+    ----------
+    database : path to sqlite database
+    **kwargs
+        Passed to Stage
+    """
+    def __init__(
+        self,
+        database,
+        output_names,
+        add_these =  ['energy', 'zenith'],
```
philippeller, line 27: could it simply be `coszen` here? Because right now you specify to add `zenith` but in the code then change this case to `coszen`, which is a bit convoluted and unexpected behaviour

philippeller:
There is an auto generated file `pisa/utils/llh_defs/poisson_gamma_mixtures.c` added, please remove it from the PR.
Also, if you could fix a few formatting issues (e.g. use black). would be appreciated.

After these changes we can merge

philippeller (changes requested):

pisa/stages/data/sqlite_loader.py
RasmusOrsoe: In principle but I would prefer not to. This dataloader is compatible only with the sqlite databases from graphnet, and we do not extract flercnn and there is not a plan to do so in the future.

pisa/stages/data/sqlite_loader.py
RasmusOrsoe: I'll rewrite this!

pisa/stages/data/sqlite_loader.py
philippeller: Sorry, "FLERCNN" is just an example. What I'm trying to say is that we could make the code more general / flexible. Instead of a boolean flag that allows for exactly two options and is fixed on the idea of comparing to "retro", we could simple here pass "retro" or "pred" or anything else like "banana" or "sharks_with_lasers" as a string. Then you can decide inside the function what happens in each case. Even assert that it should be ` in ('pred', 'retro')` right now if you want.

It's a very small detail, but I like to find ways to make stuff more general and future proof whenever possible.

RasmusOrsoe:
I've made the changes you have suggested @philippeller and tested that it works. Let me know what you think.

PR #694, opened 2022-03-07
Fixing map saving.

LeanderFischer:
The additional `os.path.join` chopped up the filename and broke the saving. Maybe I'm missing something here, but I don't see it being necessary let alone working on my end.


philippeller:
yeah, looks like a bug applying `join` twice! Thanks for the fix 👍

PR #696, opened 2022-03-09
fix bug in nusquids stage when not using taus

atrettin:
When not using taus, all probabilities from nusquids were set to zero due to the last few lines in the apply function. Sorry, I should have actually tested the stage at PR #691  in my own pipeline instead of just looking at the code... it should now cleanly switch between using taus and not using them without doing any superfluous computation or memory allocation.


BenSmithers (approved):
Everything looks good to me; things work on my end without issue. Apologies about the bug I introduced, I'll try to be more careful about that in the future!

PR #697, opened 2022-05-05
Add Bootstrap sampling stage and some small fixes

atrettin:
This PR adds a bootstrap sampling stage that can be used to estimate MC uncertainties, and a few smaller fixes:
* If a string was given for metric in the Analysis, it is now converted into a list of one string. All other functions in the class assume that `metric` is a list.
* Add `scales_as_log` to the list of parameter attributes. Otherwise, putting it into the configuration is interpreted as a new parameter called `scales_as_log`.

The `bootstrap` stage creates a bootstrap sample of the entire input data sample, i.e. it randomly draws samples with replacement until the bootstrap sample has the same size as the original sample. Instead of manhandling the entire arrays, though, it adds equivalent weights to the samples, i.e. a sample that was selected twice gets a weight of 2, and so on. 

The unit test ensures that 
* the bootstrap is repeatable with the same seed,
* the average value in each bin approximately recovers that of the original sample,
* and that the variance in each bin over bootstrap samples recovers approximately the sum of squared weights.

### DIFF
--- a/pisa/analysis/analysis.py
+++ b/pisa/analysis/analysis.py
@@ -906,6 +906,9 @@ class BasicAnalysis(object):
         
         """
 
+        if isinstance(metric, str):
+            metric = [metric]
+
         # Before starting any fit, check if we already have a perfect match between data and template
         # This can happen if using pseudodata that was generated with the nominal values for parameters
         # (which will also be the initial values in the fit) and blah...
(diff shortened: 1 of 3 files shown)


philippeller:
Hi, This PR looks clean to me and does not look like interfering with anything else. So i will merge this

PR #698, opened 2022-05-10
Make CRS2 fits reproducible

atrettin:
This PR makes fits with CRS2 (or any other NLOPT algorithm with a random component) reproducible if a seed is passed to `method_kwargs`, and adds a unit test to assert that this behaves correctly. It also adds unit tests to fluctuate `Map` and `MapSet` objects and fixes an issue where `get_random_state` without a random state would crash.


philippeller:
thanks for the fixes and improvements!

PR #700, opened 2022-05-29
Make KDE bootstrap reproducible

atrettin:
This PR makes the KDE bootstrapping that we use to estimate MC errors on KDE'd histograms reproducible. Since the `kde` module didn't support setting seeds, I had to "in-source" the bootstrapping into the KDE stage itself to do this. I also added unit tests to ensure that the histograms with errors will give the same result with the same seed and different results with different seeds.

### DIFF
--- a/pisa/stages/utils/kde.py
+++ b/pisa/stages/utils/kde.py
@@ -54,7 +54,7 @@ class kde(Stage):
         stash_hists=False,
         bootstrap=False,
         bootstrap_niter=10,
-        bootstrap_seed=0,
+        bootstrap_seed=None,
         **std_kargs,
     ):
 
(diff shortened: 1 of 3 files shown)


philippeller:
Only thing I would suggest to change, is not setting a default seed fixed to 0, but None such that by default it is random as expected. Only when you explicitly provide a seed it becomes fixed

atrettin:
> Only thing I would suggest to change, is not setting a default seed fixed to 0, but None such that by default it is random as expected. Only when you explicitly provide a seed it becomes fixed

Okay, done! I also forgot that the numeric inputs that we get when parsing a pipeline config from a text file come in as strings that have to be converted to ints first.

PR #701, opened 2022-05-31
Add missing method to assert compatibility

atrettin:
The `assert_compat` method was missing from the OneDimBinning class, and therefore the `assert_compat` method from the MultiDimBinning class would fail with an obscure error. This PR fixes this behavior and adds a unit test to ensure that the appropriate error is raised.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -36,10 +36,13 @@ jobs:
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
     - name: Test PISA example pipelines
       run: |
+        pip install pytest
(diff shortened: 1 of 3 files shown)


atrettin:
@philippeller The test is failing because the `pytest` package is not installed by default. I think that `pytest` is definitely an essential thing to have for a developer, but it may not be necessary for a use who isn't intending to run unit tests. Would it make sense to add `pytest` to the list of required packages for `develop` installs, and then also to change the GitHub workflow such that PISA is installed in developer mode for testing? Or is that too convoluted, we could also just add `pytest` to the list of packages that are always required and be done with it.

atrettin:
This should work! Shall we merge?

philippeller:
back from vacation! Yes, I'll merge this

PR #706, opened 2022-08-05
Better KDE unit tests

atrettin:
This PR adds unit testing for the "stashing" feature of the KDE to ensure that scaling stashed histograms with a subsequent weighting stage is strictly equivalent to scaling the weights first before applying the KDE. The behavior of errors is tested with both bootstrapped errors and errors set ad-hoc with the `set_variance` stage (the latter is aimed at backward compatibility with the verification sample). 

The KDE unit test also uses a much faster testing pipeline now, reducing the runtime for this unit test from 20 s to ~1 s. 

One further small tidbit is that I tracked down the origin of the `UnitStrippedWarning` one always gets when running the KDE stage and fixed the offending line. This has annoyed me for a long time.

### DIFF
--- a/pisa/utils/kde_hist.py
+++ b/pisa/utils/kde_hist.py
@@ -124,7 +124,7 @@ def get_hist(
     # the bin range for reflection
     bin_points = []
     for b in binning:
-        c = unp.nominal_values(b.weighted_centers)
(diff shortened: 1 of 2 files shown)


philippeller:
great, thank you!!

PR #707, opened 2022-10-25
bug fix on conv_llh

krish9600:
Tniy bug on conv_llh was solved and updated to return bin wise conv_llh values.


atrettin:
I understand that this fixes a bug that was found by the linter associated with a commit that was made earlier directly to master. I will merge this just to get everything working again, but please use PRs in the future.

PR #710, opened 2022-11-04
fix resetting parameter values for 2. octant fit

ellohfin:
In octant fitting, if parameter values are not reset free between the single fits, the intended behavior is that after fitting the first octant, the parameter values (except for theta_23) are reset to where the first fit started. Due to the missing deepcopy, instead, they were set to the end point of the first fit instead.


philippeller:
@ellohfin and @atrettin Is this good to merge?

atrettin:
Got the OK from Elisa that all is good, so I merged.

philippeller:
great, thank you both!

PR #713, opened 2022-12-08
Clongdouble alias

atrettin:
Just a small change that allows PISA to run on platforms other than Linux x86_64.

Somehow, there is this alias, `complex256`, that only exists on x86_64 for `clongdouble`. No idea why that would be, the workaround is also very simply to set the alias manually.

https://numpy.org/devdocs/reference/arrays.scalars.html#numpy.clongdouble

Oh, and I always get annoyed a little by this automatically generated file being shown as untracked in git, so I added that to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ pisa.egg-info
 
 # IPython notebook backups
 .ipynb_checkpoints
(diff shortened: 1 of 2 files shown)


philippeller (approved):
Looks simple enough - will merge!

PR #715, opened 2023-01-23
Adding Covariance Matrices

BenSmithers:
Basic functionality is here - hoping to continue work on this on a branch in the IceCube PISA repo rather than my personal fork. 

What's here, now
 - Adding covariance matrices to do fits with correlated nuisance parameters 
 - Code to write and read functions to and from json files 

What's not yet fully there 
 - Creating `DerivedParam`s and defining functions in config files 
 - Serialization of the full state, including `DerivedParam`s

### DIFF
--- a/pisa/core/distribution_maker.py
+++ b/pisa/core/distribution_maker.py
@@ -312,8 +312,27 @@ class DistributionMaker(object):
                         " This may cause issues.", possible_selections
                     )
 
+    def add_covariance(self,covmat):
+        """
+            Incorporates covariance between parameters. 
+            This is done by replacing relevant correlated parameters with "DerivedParams"
+                that depend on new parameters in an uncorrelated basis
+
+            The parameters are all updated, but this doesn't add the new parameters in
+            So we go to the first stage we find that has one of the original parameters and manually add this in 
+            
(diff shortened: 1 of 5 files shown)


BenSmithers:
The last two commits addressed a few bugs that I came across while using these in some fits. DerivedParams are a bit closer to serializability now

philippeller:
This looks great! So the old parameters still are usable from within each stage as if nothing happened, but internally they are represented in the rotated form, if I understood correctly?

BenSmithers:
> This looks great! So the old parameters still are usable from within each stage as if nothing happened, but internally they are represented in the rotated form, if I understood correctly?

That's right! It essentially turns the two Params into four. The original ones are replaced with `DerivedParam`s whose values are functions of the new, rotated, ones.

philippeller:
Cool! Is this PR thought as WIP, or do you want to get what you have merged?

BenSmithers:
I'd like what I have here merged!

philippeller:
cool! I then only want to run a few checks and if they pass merge it

philippeller:
Run come additional checks, it all looks good and will merge now

philippeller (approved):

PR #721, opened 2023-02-20
only change bin uncertainties if they exist

JanWeldert:
For some reason it is possible to run a distribution maker (make plots with it) without actually running the distribution maker (apply all the stages). This can lead to the fact that 'bin_unc2' is not in the containers, but the hypersurface stage tries to change it. Now the stages checks if the key exists


JanWeldert:
Again, I have no idea why all these commits appear here. The only important one is the last one

philippeller (approved):

PR #723, opened 2023-02-21
committed some checks and picked up the recent commits to master.

shiqiyugit:


pisa/core/detectors.py
```diff
@@ -30,7 +30,7 @@
 from pisa.utils.hash import hash_obj
 from pisa.utils.log import set_verbosity, logging, Levels
 from pisa.utils.random_numbers import get_random_state
-from pisa.analysis.analysis import update_param_values_detector
+#from pisa.analysis.analysis import update_param_values_detector
```
JanWeldert, line 33: If that causes problems we don't have to import the function from analysis, but then we have to define it here again. Otherwise the unit test will fail (what it currently does).

shiqiyugit: there was no issue in the past with this import loop? then I'm ok to just put it back. i took the function out from the unit test too.

shiqiyugit (commented):
revert the 2-line changes i made to prevent the import loop

pisa/core/detectors.py
JanWeldert: Hmm, I can not remember having issues with that, but I can see that it is not ideal. The problem is that when we move the function we also have to move the original update_params function out of analysis (maybe to distribution_maker.py), because we need it here.

philippeller (approved):
Without going into deep technical detail, the changes look in order to me and are mostly there to support bin masking, and a few other small improvements to the code.

philippeller:
Would still like to get @atrettin's approval too, and then we can get that PR merged

atrettin (approved):
Looks ok by me. There are many things that we already have pulled, but disentangling that would be too complicated anyway. Since unit tests are working and this has been used in analyses I trust that there aren't too glaring issues. :)

PR #727, opened 2023-03-15
Add HNL custom effective area stage - working state.

LeanderFischer:
Add custom effective area stage for OscNext HNL analysis. Can be used to re-weight single mass HNL sets for any mixing.

Uses constants (lightspeed, planck), can they be pulled from somewhere in PISA?


LeanderFischer:
Very true, will implement as such and get back to you.

philippeller:
Yep, I think like this it's very clean and modular! Will merge now

philippeller (approved):

PR #733, opened 2023-04-22
Fixed little bug in `fit_scipy`

BenSmithers:
When the `_fit_scipy` function is called with a `list` of `MapSets`, it will still try to call `data_dist.allclose(hypo_asimov_dist)`, which fails since `data_dist` is a list. This is fixed by checking if this is a list first, and if so it will then check if all the mapsets are close instead.


BenSmithers:
Since this is a minor patch, it was given its own branch. Pulling the changes from master to this patch would require another pull request, just so we can update this pull request? It's a little silly 

So... I'll just open a new patch pull request and close this one.

PR #734, opened 2023-04-24
Add ultrasurface stage and edit README accordingly.

LeanderFischer:
Add discrete systematic stage to apply ultrasurfaces. The code to predict/fit them is still in the fridge, but could potentially also be made part of pisa at some point.


atrettin (approved):
This looks all good to me, as it is basically just moving the code that already exists from the fridge into PISA. From my side, we can also just pull this for now and do the integration of the fridge scripts that actually fit the ultra surface gradients in a separate PR later.

PR #739, opened 2023-05-04
Get rid of deprecated numpy members.

LeanderFischer:
Attempting to fix unit tests.


philippeller (commented):
do we then need to bump the numpy version requirement, or is it backwards compatible?

LeanderFischer:
> 

`np.bool` was always just an alias for python built-in `bool` etc. so this should be backwards compatible afaik.

LeanderFischer:
Same holds for `np.typeDict` and `np.sctypeDict`, so that should be fine.

PR #742, opened 2023-05-23
fixed small bug in `fit_scipy`

BenSmithers:
When the `_fit_scipy` function is called with a `list` of `MapSets`, it will still try to call `data_dist.allclose(hypo_asimov_dist)`, which fails since `data_dist` is a list. This is fixed by checking if this is a list first, and if so it will then check if all the mapsets are close instead.


LeanderFischer (approved):
Looks good, will merge.

PR #743, opened 2023-05-29
Detector class bug fixes

JanWeldert:
Updating the detectors class to PISA4 didn't work as I thought. So here are some bug fixes that hopefully will resolve the issues. It is mostly about updating parameter values for the different DistributionMakers that represent the detectors.


BenSmithers:
Was it that the parameters weren't updating at all for one of the distribution makers? I might've been running into the same bug! I can look over this

pisa/core/detectors.py
```diff
@@ -147,30 +149,33 @@ def get_outputs(self, **kwargs):
...
-        for p in params.names: # now update params with det_names inside
-            for i, det_name in enumerate(self.det_names):
-                if det_name in p:
-                    cp = deepcopy(params[p])
-                    cp.name = cp.name.replace('_'+det_name, "")
-                    self._distribution_makers[i].update_params(cp)
+        if isinstance(params,Param): params = ParamSet(params)
+            
+        for distribution_maker in self.distribution_makers:
+            ps = deepcopy(params)
```
BenSmithers, line 155: Does `deepcopy` need to be used here? As I understand, it's probably to protect the internal Params whenever what's passed to this method is changed later on, but it messes with `DerivedParams` in a weird little way where they lose references to the original Params they depend on. 

I'm confident I could set it up to update those references when this copy happens, but I thought I'd check to see if there was a specific reason this was implemented as-is!

pisa/analysis/analysis.py
```diff
@@ -2518,6 +2566,8 @@ def _minimizer_callable(self, scaled_param_vals, hypo_maker, data_dist,
         scaled_param_vals = np.where(flip_x0, 1 - scaled_param_vals, scaled_param_vals)
         # Set param values from the scaled versions the minimizer works with
         hypo_maker._set_rescaled_free_params(scaled_param_vals) # pylint: disable=protected-access
+        if hypo_maker.__class__.__name__ == "Detectors":
+            update_param_values_detector(hypo_maker, hypo_maker.params.free) #updates values for ALL detectors
```
BenSmithers, line 2570: One more question - does `update_param_values_detector` need to be called every minimizer step? It looks like that could be rather expensive since it also makes calls back to `init_params`, and the free parameters should already be updated here.

LeanderFischer:
Everyone happy with this or are more eyes on this needed?

pisa/analysis/analysis.py
JanWeldert: It is needed to update the parameters for all distribution makers

philippeller (approved):

philippeller:
i think this is ready

PR #752, opened 2023-08-17
Bump python version

LeanderFischer:
We don't support python 3.7 anymore so let's bump the python version for the github workflos.


anilak41 (approved):
Looks good.

PR #753, opened 2023-08-22
Fix dependencies and update install instructions

LeanderFischer:
Nail down a few dependency version ranges, so install does not break. Also fix pip maximum version in git workflow, to be compatible with setuptools older version.


anilak41 (approved):
Looks good

PR #756, opened 2023-11-08
adding stage for external osc calc

philippeller:
This is a new PISA stage, that allows one to use an external oscillation probability calculation.
We use this in some (non IceCube) work we do. But still would be nice to merge it into master to keep things in place


LeanderFischer:
Finally had a closer look and modified it a bit. I removed _todo_ related comments, better create an issue for that, please!

Also, there is some new conflicts, since this took so long :see_no_evil:. Feel free to merge, once these are resolved and the checks pass.

LeanderFischer (approved):
See comment above.

pisa/stages/osc/layers.py
```diff
@@ -219,16 +235,24 @@ def __init__(self, prem_file, detector_depth=1., prop_height=2.):
             # w.r.t the file. The first elements of the arrays below corresponds
             # the Earth's surface, and the following numbers go deeper toward the 
             # planet's core
+            self.rhos_unweighted = self.prem[..., 1][::-1].astype(FTYPE)
```
JanWeldert, line 238: Do you really need three rhos here?

LeanderFischer:
Thanks for catching this up with `master`. Again, I'm afraid out checks are broken right now, that's why I re-triggered them. Otherwise this is ready to merge, but we need to fix the checks first :grimacing:

LeanderFischer:
Please pull from master once more, to fix the checks

PR #760, opened 2024-01-05
Re-shaped the mod_chi2's return to match the input shape

JKrishnamoorthi:
Changed the output shape of mod_chi2() function to match the input shape (shape of the map).


shiqiyugit (approved):
I dont see anything wrong but this does not seem necessary. you can always shape it in your plotting script when you call this function

PR #761, opened 2024-02-02
adding DAEMONFLUX stage

marialiubarska:
I think it's finnaly time to merge this into master...
Summary: I added daemonflux stage and a simple example notebook + cfg file (see https://arxiv.org/abs/2303.00022). Also updated flux readme as well as INSTALL.md to mention that daemonflux is now a dependency for this stage.


LeanderFischer:
Looks good to me, but I didn't check the details of the stage itself. The discrepancies that @anilak41 is seeing are confirmed with this stage and with the pure python daemonflux, right? So I assume this is unrelated to the implementation of the stage?

And apart from the notebook there is no more detailed explanations than the paper itself?

LeanderFischer (approved):
Ok, Anil confirmed that this looks fine, so I'll put it in :+1:

anilak41:
I have used this daemon flux stage and checked the DAEMON flux, which looks consistent with the one obtained using daemonflux python library.

PR #764, opened 2024-02-16
Addition of density as a free parameter for tomography

sc458930:
## Addition of density as a free parameter for tomography
-------------------------------------------
Sharmistha Chattopadhyay
--------------------------------------------
### Modified `prob3.py`
- Introduced `tomography_type`
-  `tomography_type` - [mass of earth, mass of core (with constrain), mass of core (without constrain)
- physics parameters:

1.  mass of earth (12 layers): `density_scale`
2.  mass of core - with constrain (5 layers): `core_density_scale`

### Introduced `scaling_params.py`
- Contains the scaling factors and scaling arrays corresponding to each tomography type

### Modified `layers.py`
- Introduced multiplication of scaling factor with density array

### Modified `numba_osc_tests.py`
- Introduced unit test for all the tomography types


pisa/stages/osc/prob3.py
```diff
@@ -360,12 +393,39 @@ def compute_function(self):
...
             self.nsi_params.eps_tautau = self.params.eps_tautau.value.m_as('dimensionless')
-        
         if self.neutrino_decay:
             self.decay_params.decay_alpha3 = self.params.decay_alpha3.value.m_as('eV**2')
             
         if self.lri_type is not None:
             self.lri_params.v_lri = self.params.v_lri.value.m_as('eV')
+        if self.tomography_type == "mass_of_earth":
+                self.tomography_params.density_scale = self.params.density_scale.value.m_as('dimensionless')
+                self.layers = Layers(earth_model, detector_depth, prop_height,scaling_array=self.tomography_params.density_scale)
```
JanWeldert, line 403: I think it would be better to give the `Layers` class something like a scale function that you can call here to not initialize the class every time the stage is called. That would simplify the entire code (not only this part) and should be faster.

JanWeldert (changes requested):
It would be better to modify the Layers class to be able to deal with density scale changes rather than to do everything in the osc stage and reinitialize the class every time

pisa/stages/osc/prob3.py
```diff
@@ -398,32 +395,23 @@ def compute_function(self):
...
+            elif self.tomography_type == "mass_of_core_wo_constrain":
+                    self.tomography_params.core_density_scale = self.params.core_density_scale.value.m_as('dimensionless')
+                    self.tomography_params.innermantle_density_scale = self.params.innermantle_density_scale.value.m_as('dimensionless')
+                    self.tomography_params.middlemantle_density_scale = self.params.middlemantle_density_scale.value.m_as('dimensionless')
+                    self.layers.scaling(scaling_array=self.tomography_params.scaling_factor_array)
+            self.layers.setElecFrac(self.YeI, self.YeO, self.YeM)        
+            for container in self.data:
+                self.layers.calcLayers(container['true_coszen'])
+                container['densities'] = self.layers.density.reshape((container.size, self.layers.max_layers))
+                container['distances'] = self.layers.distance.reshape((container.size, self.layers.max_layers))
```
JanWeldert, line 414: Do you also have to overwrite this? The scaling shouldn't change the distances, right?

JanWeldert (approved):
Looks better now. It might be even simpler to just multiply container['densities'] with the scaling array (and store an unscaled version of container['densities'] somewhere), but this also works.

pisa/stages/osc/prob3.py
sc458930: Yes right. We don't need to overwrite container['distances']. I have removed it. We only require till container['densities'].

PR #765, opened 2024-02-17
basic example code to get event wise information

JKrishnamoorthi:
This basic code shows the way to extract event-wise information from `DistributionMaker()`


LeanderFischer:
Hey, thanks for setting this up! Could you add some cells in between with some explanation on what is happening? Like, some information on the pipeline and why the hist stage has to be removed and so on, imagine someone very new to PISA is picking this up. Other than that, it's great, thank you!

LeanderFischer (approved):
Looking good now, thanks a lot!

PR #773, opened 2024-06-04
Deprecate duplicated prior functions

JanWeldert:
I deprecated some prior functions that just call other functions. In case of the ParamSet they will even yield wrong results since they don't know about the deamon flux stuff

### DIFF
--- a/pisa/core/param.py
+++ b/pisa/core/param.py
@@ -122,7 +122,7 @@ class Param:
     ...           range=[-10, 60]*ureg.foot, is_fixed=False, is_discrete=False)
     >>> x.value
     <Quantity(1.5, 'foot')>
-    >>> print(x.prior_llh)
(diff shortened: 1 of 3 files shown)


LeanderFischer:
Looks good to me, thanks for taking care of this!
I'm afraid our checks are currently broken, to make sure I'll quickly re-run them and see and then we can decide whether we can just merge.

JanWeldert:
Checks pass again. By the way, You can re-run the checks online without making a new commit.

LeanderFischer:
> Checks pass again. By the way, You can re-run the checks online without making a new commit.

Unfortunately, that's only true if they failed.. at least I didn't find a way to do it for successful workflows (and they were, since it's been a few weeks/months).

LeanderFischer (approved):
Looks good, thanks!

PR #774, opened 2024-06-06
Remove misleading line in fix_error.py description

finnmayhew:
Previously said: "Needed to allow mod_chi2 to behave correctly".  My understanding is that this stage reduces the accuracy of mod_chi2 by approximating MC errors throughout the space as the initial MC errors.

### DIFF
--- a/pisa/stages/utils/fix_error.py
+++ b/pisa/stages/utils/fix_error.py
@@ -1,8 +1,6 @@
 """
 Stage to take the initial errors of MC and keep them
 for all minimization.
-
-Needed to allow mod_chi2 to behave correctly
 """
 
 from __future__ import absolute_import, print_function, division


JanWeldert:
Looks good to me

JanWeldert (approved):

philippeller:
The original reason for this treatment was a bias in the nutau appearance analysis when this error was not fixed. So maybe instead we could explicitly write here that this was needed to be fixed for Dragon nutau

finnmayhew:
> The original reason for this treatment was a bias in the nutau appearance analysis when this error was not fixed. So maybe instead we could explicitly write here that this was needed to be fixed for Dragon nutau

This makes sense to me, will amend and do a new pull request.

PR #775, opened 2024-06-06
Add warning when attempting to use the analytical NSI matrix

finnmayhew:
According to an existing comment (L390 of the edited file), the analytical version of the NSI matrix is not implemented correctly.  Since this function is called in the PISA tests that run on changes to be merged, I'm writing a warning here rather than an error.


JanWeldert:
Isn't there already a warning like this?

```
[ WARNING] Warning: the analytical NSI matrix is not yet properly implemented. Instead, use the numerical NSI matrix.
[ WARNING] Evaluating analytical expressions for NSI matrix elements does not give agreement with numerical calculation! Elementwise agreement:
[[ True False  True]
 [False  True False]
 [ True False  True]]
This is expected. Going ahead with numerical calculation for now
```

PR #777, opened 2024-07-17
Raise an exeption if the (not entirely implemented) analytical versio…

finnmayhew:
…n of the effective NSI coupling matrix is accessed


finnmayhew:
Hm, running this function should probably not be in the unit tests since it's not implemented.

LeanderFischer:
> Hm, running this function should probably not be in the unit tests since it's not implemented.

Yeah, totally, but even a step before, why is this function (body) even there, if it's apparently _not implemented_? I suggest do get rid of it altogether, if there is no value in keeping the body and/or the function at all, no?

LeanderFischer:
Solved through  #786

PR #787, opened 2024-07-31
Finally fix map.plot, to be actually usable..

LeanderFischer:
Added some extra parameters to the map plotting function, which now allow:
1. using the raw bin names (from the binning config file) as title for each pid bin, by setting `pure_bin_names=True`
2. using `'layout': 'tight'` or `'layout': 'constrained'` in `fig_kw`, to produce reasonable saved figures

The second point did not work before, because the colorbar was plotted 3 times for every pid bin, and not all of them were correctly moved by the layout option.. now it is keeping track of the main axis and only plotting the colorbar once. Results are rather nice now, as you can see here:

 labeled_s_to_sqrt_b_0 3_GeV_combined_U_tau4_sq_0 0100_total

### DIFF
--- a/pisa/core/binning.py
+++ b/pisa/core/binning.py
@@ -783,7 +783,7 @@ class OneDimBinning(object):
         if self.units == ureg.dimensionless:
             units_tex = ''
         else:
-            units_tex = r' \; \left( {:~L} \right)'.format(self.units)
+            units_tex = r' \; \left[ {:~L} \right]'.format(self.units)
 
         return name_tex + units_tex
 
--- a/pisa/core/map.py
+++ b/pisa/core/map.py
@@ -577,7 +577,7 @@ class Map(object):
              outdir=None, fname=None, fmt=None, binlabel_format=None,
              binlabel_colors=["white", "black"], binlabel_color_thresh=None,
              binlabel_stripzeros=True, dpi=300, bad_color=None,
-             pure_bin_names=False
+             pure_bin_names=False, bin_id=None, full_ax=None,
              ):
         """Plot a 2D map.
 
@@ -615,7 +615,7 @@ class Map(object):
             Size of the colorbar, x-axis label, y-axis label, and title text
 
         fig_kw : mapping, optional
-            Keyword arguments passed to call to `matplotlib.pyplot.figure`;
+            Keyword arguments passed to call `matplotlib.pyplot..subplots`;
             this is only done, however, if `ax` is None and so a new figure
             needs to be created.
 
@@ -669,6 +669,11 @@ class Map(object):
             Can choose the color used for "bad" bins (e.g. NaN)
         pure_bin_names : bool, optional
             If True, use the (third dimension) bin names as they are defined in binning config, without any formatting.
+        bin_id : int, optional
+            If the map is a slice of a multi-dim map, this is the index of the slice.
+        full_ax : :class:`matplotlib.axes.Axes`, optional
+            If the map is a slice of a multi-dim map, this is the full axis object of the multi-dim map.
+
 
         Returns
         -------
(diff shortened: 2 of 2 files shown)


pisa/core/binning.py
```diff
@@ -783,7 +783,7 @@ def label(self):
         if self.units == ureg.dimensionless:
             units_tex = ''
         else:
-            units_tex = r' \; \left( {:~L} \right)'.format(self.units)
+            units_tex = r' \; \left[ {:~L} \right]'.format(self.units)
```
thehrh, line 786: What is the reason for this change? Most style guides and conventions I'm aware of don't suggest enclosing units in square brackets.

pisa/core/map.py
```diff
@@ -613,7 +615,7 @@ def plot(self, symm=False, logz=False, vmin=None, vmax=None, backend=None,
             Size of the colorbar, x-axis label, y-axis label, and title text
 
         fig_kw : mapping, optional
-            Keyword arguments passed to call to `matplotlib.pyplot.figure`;
+            Keyword arguments passed to call `matplotlib.pyplot..subplots`;
```
thehrh, line 618: "passed to call to `matplotlib.pyplot.subplots`"

pisa/core/map.py
```diff
@@ -661,7 +666,14 @@ def plot(self, symm=False, logz=False, vmin=None, vmax=None, backend=None,
         dpi : int, optional
             Dots per inch for saved figure. Default: 300
         bad_color : string, optional
-            Can choose the color used for "bad" bins (e.g. NaN) 
+            Can choose the color used for "bad" bins (e.g. NaN)
+        pure_bin_names : bool, optional
+            If True, use the (third dimension) bin names as they are defined in binning config, without any formatting.
```
thehrh, line 671: Specify default (`False`)

pisa/core/map.py
```diff
@@ -661,7 +666,14 @@ def plot(self, symm=False, logz=False, vmin=None, vmax=None, backend=None,
...
+        bin_id : int, optional
+            If the map is a slice of a multi-dim map, this is the index of the slice.
```
thehrh, line 673: Specify default and what happens then

pisa/core/map.py
```diff
@@ -661,7 +666,14 @@ def plot(self, symm=False, logz=False, vmin=None, vmax=None, backend=None,
...
+        full_ax : :class:`matplotlib.axes.Axes`, optional
+            If the map is a slice of a multi-dim map, this is the full axis object of the multi-dim map.
```
thehrh, line 675: same

pisa/core/map.py
```diff
@@ -735,15 +748,16 @@ def plot(self, symm=False, logz=False, vmin=None, vmax=None, backend=None,
...
                 _, _, pcmesh, colorbar = to_plot.plot(
                     symm=symm, logz=logz, vmin=vmin, vmax=vmax,
                     ax=small_axes[bin_idx], cmap=cmap, clabel=clabel,
                     clabelsize=clabelsize, xlabelsize=xlabelsize,
                     ylabelsize=ylabelsize, titlesize=titlesize,
-                    pcolormesh_kw=pcolormesh_kw, colorbar_kw=colorbar_kw,
+                    pcolormesh_kw=pcolormesh_kw, colorbar_kw=colorbar_kw, colorbar_label_kw=colorbar_label_kw,
                     binlabel_format=binlabel_format, binlabel_colors=["white", "black"],
                     binlabel_color_thresh=binlabel_color_thresh, binlabel_stripzeros=binlabel_stripzeros,
+                    bin_id = bin_idx, full_ax = full_ax
```
thehrh, line 760: remove spaces around `=`

pisa/core/map.py
LeanderFischer: done

pisa/core/map.py
LeanderFischer: Sure!

thehrh (approved):

Issue #45, opened 2016-06-29
Document pipeline config file


jllanfranchi:
This is extremely important to have good, full documentation. Make sure there is a concise version that covers all of the parsing features, as well as a more in-depth explanation.

PR #794, opened 2024-08-07
Apply small fixes to (mostly) docstrings of `config_parser` module and extend unit test

thehrh:
Our possibly best documentation of the expected pipeline configuration format is the module's docstring, which contains several small mistakes and is outdated or incomplete in several places (see https://icecube.github.io/pisa/pisa.utils.html#module-pisa.utils.config_parser).

There also seems to be the same small bug in two methods of `PISAConfigParser`:
* I could not find out whether the `_validate_value_types` method of `RawConfigParser` ever returned something. At the time the code was introduced, https://github.com/icecube/pisa/blob/master/pisa/utils/config_parser.py#L1184 suggests it was based on https://pypi.org/project/configparser/3.5.0/.
* In all recent versions (for example, https://github.com/python/cpython/blob/3.10/Lib/configparser.py#L1171), `_validate_value_types` does not return anything. In any case, the two methods are now identical to the same ones of  `ConfigParser` (https://github.com/python/cpython/blob/3.10/Lib/configparser.py#L1202 and https://github.com/python/cpython/blob/3.10/Lib/configparser.py#L1208).
* The only place I could find where an affected method is used is in https://github.com/icecube/pisa/blob/44ac13df4dbea3c0b12068cbd1c44234cb6f5f1d/pisa/core/pipeline.py#L738, but that script is already broken in various other ways -> now included in unit test

### DIFF
--- a/pisa/utils/config_parser.py
+++ b/pisa/utils/config_parser.py
@@ -1,14 +1,14 @@
 """
-Parse a ConfigFile object into a dict containing an item for every analysis
+Parse a config file into a dict containing an item for every analysis
 stage, that itself contains all necessary instantiation arguments/objects for
-that stage. for en example config file, please consider
+that stage. For an example config file, please consider
 :file:`$PISA/pisa_examples/resources/settings/pipeline/example.cfg`
 
 Config File Structure
 =====================
 
 A pipeline config file is expected to contain something like the following,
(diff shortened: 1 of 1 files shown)


thehrh:
See https://github.com/icecube/pisa/issues/45

JanWeldert (approved):

PR #798, opened 2024-08-08
Revive parametric Aeff stage

JanWeldert:
Bringing back the parameterized effective area stage that.

### DIFF
--- a/pisa/stages/aeff/param.py
+++ b/pisa/stages/aeff/param.py
@@ -10,23 +10,18 @@ benchmark scenarios.
 
 from __future__ import absolute_import, division
 
-from collections import OrderedDict
 from collections.abc import Mapping
 
 import numpy as np
-from scipy.interpolate import interp1d
 
 from pisa.core.stage import Stage
 from pisa.utils.fileio import from_file
-from pisa.utils.hash import hash_obj
-from pisa.utils.log import logging
-from pisa.utils import vectorizer
 from pisa.utils.profiler import profile
 
 
 __all__ = ['load_aeff_param', 'param']
 
-__author__ = 'T.C. Arlen, T. Ehrhardt, S. Wren'
+__author__ = 'T.C. Arlen, T. Ehrhardt, S. Wren, J. Weldert'
 
 __license__ = '''Copyright (c) 2014-2017, The IceCube Collaboration
  Licensed under the Apache License, Version 2.0 (the "License");
@@ -69,8 +64,8 @@ def load_aeff_param(source):
         - Callable with one argument
         - String such that `eval(val)` yields a callable with one argument
     """
-    if not (source is None or isinstance(source, (str, Mapping))):
-        raise TypeError('`source` must be string, mapping, or None')
+    if not isinstance(source, (str, Mapping)):
+        raise TypeError('`source` must be string or mapping')
 
     if isinstance(source, str):
         orig_dict = from_file(source)
@@ -83,13 +78,13 @@ def load_aeff_param(source):
     return orig_dict
 
 
-class param(Stage):
+class param(Stage): # pylint: disable=invalid-name
     """Effective area service based on parameterisation functions stored in a
     .json file.
-    Transforms an input map of a flux of a given flavour into maps of
-    event rates for the two types of weak current (charged or neutral),
-    according to energy and cosine zenith dependent effective areas specified
-    by parameterisation functions.
+    Transforms an input map of a flux of a given flavour (and interaction)
+    into maps of event rates, according to energy and cosine zenith dependent 
+    effective areas specified by parameterisation functions.
+    Requires true_energy, true_coszen, and weights to be present in the container.
     Parameters
     ----------
     params : ParamSet
@@ -106,27 +101,23 @@ class param(Stage):
         expected_params = (
             'aeff_energy_paramfile',
             'aeff_coszen_paramfile',
-            'livetime', 
+            'livetime',
             'aeff_scale'
         )
 
-        super(self.__class__, self).__init__(
+        super().__init__(
             expected_params=expected_params,
             **std_kwargs,
         )
-        
-        energy_param_source = self.params.aeff_energy_paramfile.value
-        coszen_param_source = self.params.aeff_coszen_paramfile.value
-        
-        self.energy_param = load_aeff_param(energy_param_source)
-        self.coszen_param = load_aeff_param(coszen_param_source)
+
+        self.energy_param = load_aeff_param(self.params.aeff_energy_paramfile.value)
+        self.coszen_param = load_aeff_param(self.params.aeff_coszen_paramfile.value)
 
     @profile
     def apply_function(self):
-        # read out
         aeff_scale = self.params.aeff_scale.m_as('dimensionless')
         livetime_s = self.params.livetime.m_as('sec')
-        
+
         for container in self.data:
             scale = aeff_scale * livetime_s * np.ones(container.size)
             if container.name in self.energy_param.keys():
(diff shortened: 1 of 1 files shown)


pisa/stages/aeff/param.py
```diff
@@ -0,0 +1,140 @@
...
+            'livetime', 
+            'aeff_scale'
+        )
+
+        super(self.__class__, self).__init__(
+            expected_params=expected_params,
+            **std_kwargs,
+        )
+        
+        energy_param_source = self.params.aeff_energy_paramfile.value
```
thehrh, line 118: unnecessary variable, pass to `load_aeff_param` directly

pisa/stages/aeff/param.py
```diff
@@ -0,0 +1,140 @@
...
+        coszen_param_source = self.params.aeff_coszen_paramfile.value
```
thehrh, line 119: same

pisa/stages/aeff/param.py
```diff
@@ -0,0 +1,140 @@
...
+        aeff_scale
+    """
+    def __init__(
+        self,
+        **std_kwargs,
+    ):
+        expected_params = (
+            'aeff_energy_paramfile',
+            'aeff_coszen_paramfile',
+            'livetime',
```
thehrh, line 109: trailing whitespace (see pylint), also in lines 117, 120, 129

pisa/stages/aeff/param.py
```diff
@@ -0,0 +1,140 @@
...
+        
+        energy_param_source = self.params.aeff_energy_paramfile.value
+        coszen_param_source = self.params.aeff_coszen_paramfile.value
+        
+        self.energy_param = load_aeff_param(energy_param_source)
+        self.coszen_param = load_aeff_param(coszen_param_source)
+
+    @profile
+    def apply_function(self):
+        # read out
```
thehrh, line 126: useless comment

pisa/stages/aeff/param.py
```diff
@@ -0,0 +1,140 @@
...
+from pisa.utils.fileio import from_file
+from pisa.utils.hash import hash_obj
+from pisa.utils.log import logging
+from pisa.utils import vectorizer
+from pisa.utils.profiler import profile
+
+
+__all__ = ['load_aeff_param', 'param']
+
+__author__ = 'T.C. Arlen, T. Ehrhardt, S. Wren'
```
thehrh, line 29: include yourself

pisa/stages/aeff/param.py
```diff
@@ -0,0 +1,140 @@
...
+        **std_kwargs,
+    ):
+        expected_params = (
+            'aeff_energy_paramfile',
+            'aeff_coszen_paramfile',
+            'livetime', 
+            'aeff_scale'
+        )
+
+        super(self.__class__, self).__init__(
```
thehrh, line 113: `super().__init__` (https://realpython.com/python-super/#a-super-deep-dive)

pisa/stages/aeff/param.py
```diff
@@ -0,0 +1,140 @@
...
+            **std_kwargs,
+        )
+        
+        energy_param_source = self.params.aeff_energy_paramfile.value
+        coszen_param_source = self.params.aeff_coszen_paramfile.value
+        
+        self.energy_param = load_aeff_param(energy_param_source)
+        self.coszen_param = load_aeff_param(coszen_param_source)
+
+    @profile
```
thehrh, line 124: Is it encouraged to decorate `apply_function` like this, or rather to make use of the mechanism in https://github.com/icecube/pisa/blob/master/pisa/core/stage.py#L363?

pisa/stages/aeff/param.py
```diff
@@ -0,0 +1,140 @@
...
+            ...
+        }
+    Note that the `transform_groups` (container.name) defined
+    in a pipeline config file using this must match the groupings defined
+    above.
+    `val`s can be one of the following:
+        - Callable with one argument
+        - String such that `eval(val)` yields a callable with one argument
+    """
+    if not (source is None or isinstance(source, (str, Mapping))):
```
thehrh, line 72: Why allow a `source` of `None` here? This will just raise a `TypeError` in l. 80 below.

pisa/stages/aeff/param.py
```diff
@@ -0,0 +1,140 @@
...
+        raise TypeError('`source` must be string, mapping, or None')
```
thehrh, line 73: same

pisa/stages/aeff/param.py
```diff
@@ -0,0 +1,140 @@
...
+                        % type(source))
+
+    return orig_dict
+
+
+class param(Stage):
+    """Effective area service based on parameterisation functions stored in a
+    .json file.
+    Transforms an input map of a flux of a given flavour into maps of
+    event rates for the two types of weak current (charged or neutral),
```
thehrh, line 90: Seems to specific. What types of events are modified depends on keys in param files. Also, the service doesn't convert an input type of events into another type, does it, but simply scales weights of event types that are already there?

pisa/stages/aeff/param.py
```diff
@@ -0,0 +1,140 @@
...
+    according to energy and cosine zenith dependent effective areas specified
+    by parameterisation functions.
```
thehrh, line 92: Document here that the existence of keys `true_energy`, `true_coszen`, and `weights` is required in containers?

pisa/stages/aeff/param.py
JanWeldert: Good question. I have seen the @profile multiple times in the code, not sure what would be the better option.

pisa/stages/aeff/param.py
JanWeldert: Correct

pisa/stages/aeff/param.py
JanWeldert: I can add that, but I think without that information you wouldn't make it to this stage.

pisa/stages/aeff/param.py
thehrh: Here, if I'm not mistaken, whenever you run a pipeline with verbosity set to debug level, timing will be reported for the apply function... In the other case, since the mechanism applies to the base class itself and to setup, compute, and apply functions, we don't have to rely on decorators being present in each service to get timing information.

thehrh: Hence, I suggest removing the decorator if you have no strong reason for including it.

thehrh (approved):

thehrh:
Note that we have VLVNT parameterisations in https://github.com/icecube/pisa/tree/master/pisa_examples/resources/aeff, and that there are lists of discrete energies and aeffs that need to be interpolated in there.

thehrh (commented):
Just address the interp1d bounds question, then we can merge

pisa/stages/aeff/param.py
```diff
@@ -63,19 +64,61 @@ def load_aeff_param(source):
...
+                raise ValueError('No effective area values are provided for %s'%(k))
+            if not (is_energy or is_coszen):
+                raise ValueError('No energy or coszen values are provided for %s'%(k))
+
+            var = 'energy' if is_energy else 'coszen'
+            x_vals = func[var]
+            aeff_vals = func['aeff']
+
+            param_func = interp1d(x_vals, aeff_vals, kind='linear',
+                                  bounds_error=False, fill_value=0)
```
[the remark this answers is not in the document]
JanWeldert: The advantage here is that you can use the same flux/events file with different effective area ranges.

pisa/stages/aeff/param.py
```diff
@@ -13,6 +13,7 @@
 from collections.abc import Mapping
 
 import numpy as np
+from scipy.interpolate import interp1d
```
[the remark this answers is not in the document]
JanWeldert: At the moment we limit scipy to a version where interp1d is still present, but I can create an issue.

Issue #779, opened 2024-07-22
pisa currently incompatible with python 3.11 and 3.12


thehrh:
One might be inclined to use the above python releases because [the installation guide](https://github.com/icecube/pisa/blob/master/INSTALL.md) states that "you can choose your preferred version of python >= 3.10". This is followed by the example shell command `conda create -n pisa_env python=3.10`.

However, as it stands, the installation fails under python 3.12 and python 3.11. This is probably most easily avoided right now by removing the ambiguity in the python version requirements. In any case, below are some currently existing incompatibilities that prevent using each version (there are probably more).

**python 3.12**
- [ `versioneer.py`](https://github.com/icecube/pisa/blob/master/versioneer.py) is too old (see https://docs.python.org/3/whatsnew/3.12.html#configparser). Is there a reason why we can't ship a newer version of [versioneer](https://github.com/python-versioneer/python-versioneer/blob/master/README.md) which supports python 3.12 as well?
- numpy only supports python 3.12 starting with [v1.26.0](https://numpy.org/doc/stable/release/1.26.0-notes.html), but we require < 1.23

**python 3.11**
* pip seems to be trying to build [scikit-learn <= 1.1.2](https://pypi.org/project/scikit-learn/1.1.2/#files) from its source distribution because there are no wheels for python 3.11 (only for scikit-learn >= 1.1.3), and I'm getting some `Cython.Compiler.Errors.CompileError` apparently when some extensions are being compiled
  * there is no conda package for python 3.11 either (check https://anaconda.org/conda-forge/scikit-learn/files?version=1.1.2 for example)
* numpy only supports python 3.11 starting with [v1.23.2](https://numpy.org/doc/stable/release/1.23.2-notes.html), but we require < 1.23

LeanderFischer:
Hey, thanks a lot for the detailed issue! Yes, I also saw that those install instructions were a bit imprecise, they were very [recently set up](https://github.com/icecube/pisa/commit/12faa1c585fff357327705c4377b08c7a261d589) by some newer collaborators.. Making the requirements more specific sounds like a very good idea to stop people from struggling with the installation.

In any case, if you are building in an environment where `cvmfs` is available, I suggest do use the instructions I've set up and tested for a [specific python version](https://github.com/icecube/pisa/blob/master/INSTALL.md#instructions-to-install-pisa-on-madison-working-group-servers-cobalts-current-guide-from-august-2023) there.

thehrh:
[...]

PR #806, opened 2024-08-15
Update to latest release of versioneer (0.29)

thehrh:
Expected to be Python 3.11 and 3.12 compatible.

Some tests done after updating versioneer (see https://github.com/icecube/pisa/issues/779#issuecomment-2289994626):
* successful installation of PISA with Python 3.10.14, version number returned by [`pisa.__version__`](https://github.com/icecube/pisa/blob/master/pisa/__init__.py#L84) agrees with version in package metadata (pisa.egg-info/PKG-INFO)
* using Python 3.12.5, I was able to execute the updated copy of `versioneer.py` (0.29) in the source tree (`python versioneer.py setup`), and also perform the [vendored mode install](https://github.com/python-versioneer/python-versioneer?tab=readme-ov-file#vendored-mode) using the pip-installed versioneer package (also 0.29)

Not explicitly tested (maybe someone else could verify by trying something similar to the above using this branch):
* Python 3.8 compatibility (according to https://github.com/python-versioneer/python-versioneer?tab=readme-ov-file#the-versioneer, it is also supported)


thehrh:
Odd that the builds now fail. I sometimes get the exact same problem which I can prevent via `export CFLAGS=-I<whatever numpy.get_include() returns>` before running the setup (which I thought would be addressed by https://github.com/icecube/pisa/blob/master/setup.py#L217C34-L217C53). It might have to do with https://github.com/icecube/pisa/blob/master/setup.py#L283.

PR #813, opened 2024-08-22
Allow for ultrasurfaces fit results file path relative to pisa resources search paths

### LINKED ISSUE #778: Ultrasurfaces expect absolute path to fit results file
In [these lines](https://github.com/icecube/pisa/blob/15eed8557db13638ff241cfbaee8a5ca779f175a/pisa/stages/discr_sys/ultrasurfaces.py#L136-L137), the US fit results are loaded, but this expects an absolute path (so not generic for different clusters/wgs. This could simply be replaced with the [find_resource](https://github.com/icecube/pisa/blob/master/pisa/utils/resources.py#L38) function as it is done for the [flux stages](https://github.com/icecube/pisa/blob/master/pisa/stages/flux/mceq_barr.py#L245-L249). Should be a very quick fix.

thehrh:
Closes #778

### DIFF
--- a/pisa/stages/discr_sys/ultrasurfaces.py
+++ b/pisa/stages/discr_sys/ultrasurfaces.py
@@ -11,6 +11,7 @@ from pisa import FTYPE
 from pisa.core.stage import Stage
 from pisa.utils.log import logging
 from pisa.utils.profiler import profile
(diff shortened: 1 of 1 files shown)


LeanderFischer (approved):
Awesome, thanks for the quick fix!

Issue #795, opened 2024-08-08
Missing guide on how to create a service


thehrh:
This should be a part of general [PISA conventions](https://github.com/icecube/pisa/blob/master/pisa/general_conventions.md) (again).

The short readme https://github.com/icecube/pisa/blob/master/pisa/stages/README.md contains some useful bits and pieces but is far from comprehensive.

See also https://github.com/icecube/pisa/issues/513.

thehrh:
We are also missing a generic approach to testing a given service. The suggestion to "look at the `pisa.core.pipeline.stage_test()` function to implement tests for your service." from [general conventions](https://github.com/icecube/pisa/blob/cfeb9b45a38f981059197c293f1a53aa5ab81a39/pisa/general_conventions.md) is clearly not of any help.

thehrh:
included in #817

PR #817, opened 2024-08-27
Service transparency and usability: expected data keys, supported representations, unit tests

### LINKED ISSUE #809: Let each `Stage` transparently declare expected and new container keys upon initialisation
Currently, it is not exactly obvious to the user which sequence of stage.services can be combined into one working pipeline configuration without digging through a lot of code.

While each `Stage` has an [`expected_params`](https://github.com/icecube/pisa/blob/a752eaf0f7d75c3d002d3c8ccde4a100ca1bf594/pisa/core/stage.py#L88) attribute in order to communicate which params need to be defined, there is no analogous mechanism for the data fields/container keys based on which a given service makes its computations, and, e.g., modifies weights. These are generally not even documented, in contrast to `expected_params`. Without all of `expected_params` for each stage.service, it won't initialise, and neither will a pipeline containing it.

As an example (just one of many), take [osc.prob3](https://github.com/icecube/pisa/blob/a752eaf0f7d75c3d002d3c8ccde4a100ca1bf594/pisa/stages/osc/prob3.py), which requires each container to contain the keys [`"true_coszen"`](https://github.com/icecube/pisa/blob/a752eaf0f7d75c3d002d3c8ccde4a100ca1bf594/pisa/stages/osc/prob3.py#L344) and [`"true_energy"`](https://github.com/icecube/pisa/blob/a752eaf0f7d75c3d002d3c8ccde4a100ca1bf594/pisa/stages/osc/prob3.py#L449). As of now, this is completely undocumented, and it is up to the service itself to deal with missing keys. However, shouldn't checking for missing ones be generically performed by the base `Stage` instead (compare [expected params check](https://github.com/icecube/pisa/blob/a752eaf0f7d75c3d002d3c8ccde4a100ca1bf594/pisa/core/stage.py#L114))?
A simple example is [utils.fix_error](https://github.com/icecube/pisa/blob/a752eaf0f7d75c3d002d3c8ccde4a100ca1bf594/pisa/stages/utils/fix_error.py), which requires the presence of [`"errors"`](https://github.com/icecube/pisa/blob/a752eaf0f7d75c3d002d3c8ccde4a100ca1bf594/pisa/stages/utils/fix_error.py#L37) during `compute` (undocumented, no exception handling beyond [that in `Container` itself](https://github.com/icecube/pisa/blob/a752eaf0f7d75c3d002d3c8ccde4a100ca1bf594/pisa/core/container.py#L441)).

Of course, while typically a given data loader, during `setup`, populates (most) keys required by any possible later stage.service, other services may also add keys on which some subsequent service depends, and not necessarily during `setup` (for instance, `utils.hist` adds the previously mentioned `"errors"` key upon [`apply`](https://github.com/icecube/pisa/blob/a752eaf0f7d75c3d002d3c8ccde4a100ca1bf594/pisa/stages/utils/hist.py#L112)).

If a given service declared any added/new keys during initialisation, however, on the one hand it would become apparent that `utils.fix_error` could follow `utils.hist`, for example, and on the other we should be able to detect an inconsistent pipeline config upon pipeline initialisation already.

Through the above, it should also in principle become possible to create a default unit test with no additional work for anyone implementing a service, by producing dummy inputs with the required keys (perhaps combined with some rudimentary detection of the domain of definition for each required key, e.g., energy >= 0 GeV, or -1 <= coszen <= 1).

Status of development in https://github.com/thehrh/pisa-1/tree/stage_declare_expected_keys.

Would also close #811

### LINKED ISSUE #811: `ContainerSet`'s `_representation` attribute is not set upon initialisation (possibly related to unused `data_specs` constructor parameter)
Trying to access the [`representation` attribute](https://github.com/icecube/pisa/blob/a752eaf0f7d75c3d002d3c8ccde4a100ca1bf594/pisa/core/container.py#L62) of a newly instantiated `ContainerSet` results in:

  `AttributeError: 'ContainerSet' object has no attribute '_representation'. Did you mean: 'representation'?`

because `self._representation` is [not assigned a default value](https://github.com/icecube/pisa/blob/a752eaf0f7d75c3d002d3c8ccde4a100ca1bf594/pisa/core/container.py#L39) . This does not seem like intended behaviour.

Possible solution: replace the constructor parameter `data_specs` (which is unused right now, bug!) with `representation`, and call the [representation setter](https://github.com/icecube/pisa/blob/a752eaf0f7d75c3d002d3c8ccde4a100ca1bf594/pisa/core/container.py#L66) with that argument?

### LINKED ISSUE #814: Bugs in `reco.simple_param` service
For documentation before making a PR...

Generic service test revealed a few bugs:
* https://github.com/icecube/pisa/blob/70f71538fc5eadccb3419fae4a27a24781d75cdb/pisa/stages/reco/simple_param.py#L70: odict keys/values are not subscriptable (`list(matches.keys()/values())[0]`)
* https://github.com/icecube/pisa/blob/70f71538fc5eadccb3419fae4a27a24781d75cdb/pisa/stages/reco/simple_param.py#L456, https://github.com/icecube/pisa/blob/70f71538fc5eadccb3419fae4a27a24781d75cdb/pisa/stages/reco/simple_param.py#L480, https://github.com/icecube/pisa/blob/70f71538fc5eadccb3419fae4a27a24781d75cdb/pisa/stages/reco/simple_param.py#L505: now all have to read `in container:`  -> `in container.keys`

thehrh:
- draft PR to enhance visibility of foreseen changes (cf. #809)

- main changes to `Stage` base class:
  - new `expected_container_keys` attribute must be set (no deviation from current behaviour if empty/wrong, but unit test likely to fail)
  - new `supported_reps` attribute may be set (informs new setters for `calc_mode` and `apply_mode`)
  - new boolean `in_standalone_mode` attribute (when set to True: automatically rerun `setup()` with reset `data` when `calc_mode` is changed, but not when set to False, in which case user must take care to rerun pipeline setup manually, see [pisa modes guide notebook](https://github.com/icecube/pisa/blob/39abb5c8c72f0c52bd603baaf4408628ffa96f92/pisa_examples/pisa_modes.ipynb)
- service unit tests:
  - performed by `pisa_tests/test_services.py`, which is modelled after `run_unit_tests.py`
  - executed as part of workflow `.github/workflows/pythonpackage.yml` (each test needs to pass)
  - if given service has non-standard args, requires an `init_test` function that returns a test instance
  - creates generic (not necessarily physical) toy data with few special cases, tries to set `calc_mode` and `apply_mode`, then calls `setup()` and `run()`

- resolves #809, resolves #811, resolves #814


thehrh:
resolves #795

JanWeldert:
The formatting changes were done by Thomas manually. So no automatic tool that can screw up our code.
It took some time but now all stages (except for osc.external' and 'flux.airs' which are ignored) pass their test. That doesn't necessarily mean it makes sense (from a physics point of view) what they do but their implementation works and it is easy to run them with some dummy examples. At the end, this is all we check here.

philippeller (approved):
Ok, sounds reasonable! From my side green light 🚥

Issue #804, opened 2024-08-13
Supporting Bayesian Analysis


JanWeldert:
At the moment we focus on the frequentist approach. To also support Bayesian analyses we would mainly need some sort of MCMC sampling tool and its integration into the analysis class. Examples how to perform a Bayesian analysis would also be useful.

JoshHenryPeterson:
I talked with some MSU people and they pointed me towards some work on implementing MCMC.  Apparently they were able to generate a sensitivity?  Their code lives in the fridge here:  analysis/DRAGON_NSI_3y/scripts/sampling.  It's a bit old so there may be some technical debt but it could be a good starting point.

PR #829, opened 2024-09-27
First version of an MCMC sampler

JanWeldert:
First version of an simple MCMC sampler as part of the analysis class. You need to manually install emcee at the moment.

This PR is part of #804.

I also sneaked in a fix for the search for shared parameters in the detectors class.

_(Again ignore the first 7 commits)_


pisa/analysis/analysis.py
```diff
@@ -2681,6 +2681,83 @@ def _minimizer_callback(self, xk, **unused_kwargs): # pylint: disable=unused-arg
...
+
+        """
+        import emcee
+
+        params_truth = hypo_maker.params.free
+        x0 = np.array([p._rescaled_value for p in params_truth])
+        bounds = np.repeat([[0,1]], len(x0), axis=0)
+
+        ndim = len(hypo_maker.params.free)
+        p0 = np.random.rand(ndim * nwalkers).reshape((nwalkers, ndim))
```
thehrh, line 2720: should be made reproducible (using random state from utils.random_numbers)

pisa/analysis/analysis.py
```diff
@@ -2681,6 +2681,83 @@ def _minimizer_callback(self, xk, **unused_kwargs): # pylint: disable=unused-arg
...
+        
+        def func(scaled_param_vals, bounds, data_dist, hypo_maker, metric):
```
thehrh, line 2722: give it a very brief docstring for now at least

thehrh (changes requested):
see inline

philippeller:
When the used metric is chi2, it should be halved so that it can be interpreted as a makeshift log-likelihood. I don't see that happening in the code. Can you verify you get compatible results using chi2 and llh? And does the result also give compatible results with our standard analysis?

JanWeldert:
Ah right, good point. I will add the check for chi2 metrics.
I haven't compared it to any frequentist analysis so far, mainly because it is pretty slow.

JanWeldert:
BTW, I have a little toy example comparing the Frequentist and Bayesian results. I won't put it into pisa, but maybe can go into the fridge?

pisa/analysis/analysis.py
```diff
@@ -2681,6 +2682,136 @@ def _minimizer_callback(self, xk, **unused_kwargs): # pylint: disable=unused-arg
...
+            Array containing all points in the parameter space visited by each walker.
+            It is sorted by steps, so all the first steps of all walkers come first.
+            To for example get all values of the Nth parameter and the ith walker, use
+            scaled_chain[i::nwalkers, N].
+
+        scaled_chain_burnin : numpy array (optional)
+            Same as scaled_chain, but for the burn in phase.
+
+        """
+        import emcee
```
JoshHenryPeterson, line 2734: May want to add a comment specifying which sampling algorithm the emcee package uses, for future reference.

pisa/analysis/analysis.py
```diff
@@ -2681,6 +2682,136 @@ def _minimizer_callback(self, xk, **unused_kwargs): # pylint: disable=unused-arg
...
+
+        assert 'llh' in metric or 'chi2' in metric, 'Use either a llh or chi2 metric'
+        if 'chi2' in metric:
+            warnings.warn("You are using a chi2 metric for the MCMC sampling."
+                          "The sampler will assume that llh=0.5*chi2.")
+
+        params_truth = hypo_maker.params.free
+        x0 = np.array([p._rescaled_value for p in params_truth])
```
JoshHenryPeterson, line 2742: As far as I can tell, x0 is only used in the following line, and we only seem to use the length of the array.  May be able to save some time and memory by extracting the length directly from params_truth.

JoshHenryPeterson (changes requested):

JoshHenryPeterson:
If you are going to put a toy example in the fridge please put it somewhere where it can easily be found, like maybe the bootcamp directory?  I had to dig pretty deep to find the previous bayesian sampling work and I don't want that to happen to future analyzers.

pisa/analysis/analysis.py
JanWeldert: I added the option to define the sampling algorithm and pointed to the emcee docs where one can find the different options and the default

pisa/analysis/analysis.py
JanWeldert: Good point, I simplified the code a bit

JanWeldert:
Could also have its own MCMC sampling or Bayesian analysis example folder

JoshHenryPeterson (approved):

thehrh (approved):

PR #835, opened 2024-11-10
Introducing support for variable binning with event classes (species)

marialiubarska:
!!!DO NOT MERGE UNTIL APPROVED BY REQUESTED REVIEWERS!!!

Summary of the changes:
1. Introduced a new binning class, `VarMultiDimBinning`, which holds a list of `EventSpecie`s
2. each `EventSpecie` is defined by name, binning and selection string (can also be None)
3. Added support for `VarMultiDimBinning` into `Map`, so can support arithmetic operations, hashing, eq/neq, metric calculation
4. Limited support in `Container`, no translation from/to `VarMultiDimBinning` shape (planning to add translation `VarMultiDimBinning <-> 'events'`, translation to/from MultidimBinning isn't necessary imho, but this is up for debate)
5. Modified `utils.hist` stage by adding handling of `VarMultiDimBinning`. This is where sample is divided based on selection. For now I used if/else statements, but it leads to some code duplication, so I'm open for suggestion (maybe we just make two hist stages, with and without `VarMultiDimBinning` support?)

Main idea is that analysis functionality should be minimally affected (I think only `get_detailed_metric_info` should really be affected), this is the reason for introducing this at binning level. However I haven't tested all functionality yet

I have tested running Pipeline with identical 4d binning, one using `MultiDimBinning` and one with `VarMultiDimBinning` containing a single `EventSpecie` with the same 4d binning and `selection=None`. Resulting histograms are identical. You can see results of the timed test below:
image

Main TODOs so far:
1. test with a full analysis to make sure nothing breaks
2. in hist stage add a check to make sure selected event species don't overlap - we don't want double counting
3. add example notebook and missing docstrings
4. move selection ,ask calculations to setup_function

Let me know if you have any suggestions!


JanWeldert:
Thanks for implementing this. I will give it a more detailed look, but have one first comment. I think we need a translation from each binning mode (events, multi dim, and var binning) to each other because we want to be able to run the stage before the hist stage in event as well as binned mode. We can always translate to event mode first and then change the binning mode though

thehrh:
In general, I'm a bit worried about adding another "events" class (here `EventSpecie`, whose main reason for existence seems to be to wrap a `MultiDimBinning` instance), cf `core.events` and `core.events_pi`. For example, `EventsPi` hold metadata beyond a simple selection string.

Does one actually need an `EventSpecie` class? Why not directly store a list of `MultiDimBinnings`, metadata, event names etc. in `VarMultiDimBinning`, instead of taking the detour via `EventSpecie`?

JanWeldert:
Now that #849 is merged, can we close this?

thehrh:
Some features and ideas from here made it into #849.

Issue #834, opened 2024-11-08
Support user-defined (in)equality constraints also for scipy minimisation


thehrh:
The absence of constraints (inequality or equality) is hard-coded by PISA within `_fit_scipy` , see https://github.com/icecube/pisa/blob/41c0e89cd48c2529cc9d586ea286f4a1a6fdd048/pisa/analysis/analysis.py#L1726 and https://github.com/icecube/pisa/blob/41c0e89cd48c2529cc9d586ea286f4a1a6fdd048/pisa/analysis/analysis.py#L1826, even though e.g. SLSQP, L-BFGS-B, or SHGO support them.

According to the documentation, inequality constraints for NLOPT can be defined as part of minimiser settings (https://github.com/icecube/pisa/blob/41c0e89cd48c2529cc9d586ea286f4a1a6fdd048/pisa/analysis/analysis.py#L848), and are converted into a callable internally (https://github.com/icecube/pisa/blob/41c0e89cd48c2529cc9d586ea286f4a1a6fdd048/pisa/analysis/analysis.py#L2446).

PR #841, opened 2024-12-05
Enable (in)equality constraints for use with local and global scipy minimizers

thehrh:
* also (re)introduce support for a few more scipy minimisers
* redefine `_minimizer_callback` signature to support "trust-constr" minimiser
* add a few quick minimiser tests (with and without constraints)
* minor touchups

### DIFF
--- a/pisa/analysis/analysis.py
+++ b/pisa/analysis/analysis.py
@@ -121,6 +121,20 @@ def scipy_constraints_to_callables(constr_dicts, hypo_maker):
         cd["fun"] = partial(constr_func, constr_func_params=constr_func_params)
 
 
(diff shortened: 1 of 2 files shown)


thehrh:
closes #834

thehrh:
still need to make sure that constraints are correctly passed through to other minimisation schemes apart from simple local ones

thehrh:
closes #842

thehrh:
Note that various of the new minimiser unit tests "fail" (error is caught and reported but not re-raised) due to scipy bugs/quirks. I think pisa shouldn't prevent the user from trying out the various minimisation interfaces and options scipy provides in principle, even if some of these are buggy (either always or only when given certain configurations). This should also lower maintenance cost and raise awareness of existing enhancement potential.

JanWeldert (approved):

PR #845, opened 2025-01-13
Update INSTALL.md

bleaktwig:
Cleared up the python version requirements when installing pisa through CVMFS and venv.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -112,11 +112,8 @@ Load the CVMFS environment:<br>
 ```bash
 unset OS_ARCH; eval `/cvmfs/icecube.opensciencegrid.org/py3-v4.2.1/setup.sh`
 ```
-<details>
-  <summary>on one of IceCube's Cobalt nodes</summary>
-   
-   Verify that `which python` outputs `/cvmfs/icecube.opensciencegrid.org/py3-v4.2.1/RHEL_7_x86_64/bin/python`.
-</details>
+> [!WARNING]
+> Make sure you are using python version between 3.8 and 3.10, otherwise the installation will fail due to the `setuptools` package version.
 
 ```bash
 python -m venv /PATH/TO/<VENV NAME>


LeanderFischer (approved):
Nice, looks good to me.

thehrh:
Can someone explain?

bleaktwig:
> Can someone explain?

In the Zeuthen wg server the default python version is 3.11, which breaks the pisa installation. Since the required python version is specified at the end of the `INSTALL.md` file, I thought it was a good idea to add this small warning so people don't accidentally try to install pisa with python 3.11 (which happened to me twice).

thehrh:
The cvmfs method we support via these installation instructions specifically ensures that the py3-v4.2.1 cvmfs distribution is used. Why or how does your local WG server's default python version come into play?

Do you agree that
* python version requirements are not specific to the cvmfs + venv install,
* that the method we support, using the py3-v4.2.1 distribution, ensures the use of Python 3.10.2,
* and that in this case the given output of `which python` is correct?

bleaktwig:
Ok, sorry, I was missing some context in my explanation. The Zeuthen wg server uses RHEL 9, so the py3-v4.2.1 cvmfs distribution cannot be used since its only available up to RHEL 8. I can do a more thorough edit of the md file to explain how to install using py3-v4.2.1 and py3-v4.3.0.

thehrh:
I see. The instructions aren't meant to solve problems that arise from deviating from the supported method(s), even if you are obviously forced to do so here. (We used to have a wiki for any unofficial (for example site-specific) instructions, which didn't have to undergo review.) 

Could you present your case and the solution you found for RHEL 9/py3-v4.3.0 in the developers channel on slack? That would allow us to jointly discuss whether this should replace the currently supported cmvfs installation method or whether it should be part of some less formal piece of documentation (for now).

thehrh:
Closing this as per slack discussion and given upcoming work on #779

PR #851, opened 2025-04-04
Make `VarBinning` accept cut variables that are simultaneously part of a `MultiDimBinning`

thehrh:
For context, see last osc. WG call. The corresponding cut expressions could now look as follows:
```
binning.split = (E_reco < 10) & (pid >= 0.4) & (pid < 0.64) , (E_reco >= 10) & (Y < 0.38) & (pid >= 0.4) & (pid < 0.64) , (E_reco >= 10) & (Y >= 0.38) & (pid >= 0.4) & (pid < 0.64), (E_reco < 30) & (pid >= 0.64), (E_reco >= 30) & (Y < 0.32) & (pid >= 0.64), (E_reco >= 30) & (Y >= 0.32) & (pid >= 0.64)
```

### DIFF
--- a/pisa/core/binning.py
+++ b/pisa/core/binning.py
@@ -3070,10 +3070,9 @@ class VarBinning():
 
     If a `OneDimBinning` is passed to `selections`, the corresponding
     dimension must not be a dimension of any `MultiDimBinning` in `binnings`.
-    Similarly, if a list of cut expressions is passed, none of its contained
-    variables may simultaneously serve as a `MultiDimBinning`'s dimension.
-    While such scenarios needn't always be dangerous, there are probably less
-    convoluted ways of achieving them.
+    If a list of cut expressions is passed instead, its contained
+    variables may simultaneously serve as a `MultiDimBinning`'s dimension
+    (not necessarily dangerous as long as you know what you are doing).
 
     A warning is issued when `selections` is of type `OneDimBinning` and
     all `binnings` entries are identical.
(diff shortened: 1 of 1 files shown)


pisa/core/binning.py
```diff
@@ -3091,11 +3090,12 @@ def __init__(self, binnings, selections):
...
-            # determine whether any variable is part of this binning
-            invalid_sel_vars = self._selection_vars_in_mdb(b, selections)
-            if invalid_sel_vars:
+            # Determine whether any variable is part of this binning
+            # and raise error when selections are given as OneDimBinning
+            sel_vars_in_mdb = self._selection_vars_in_mdb(b, selections)
+            if sel_vars_in_mdb and isinstance(selections, OneDimBinning):
                 raise ValueError(
-                    f"Selection variables {invalid_sel_vars} may not"
+                    f"Selection variables {sel_vars_in_mdb} may not"
```
JoshHenryPeterson, line 3098: It would be good if this error message specified that the selection variables given as OneDimBinning cannot be a part of any MultiDimBinning.  This could help with future debugging.

JanWeldert: Isn't that exactly what the error message does?

JanWeldert (approved):

JanWeldert:
Just need to fix the failed tests

pisa/core/binning.py
JoshHenryPeterson: Not quite.  The error message is unchanged from what it was before and doesn't include any info about OneDimBinning being able to throw the error.  Ultimately though this is a minor critique and I think the pull request can go through with or without changing the error message.

JanWeldert: Oh, you mean the message should explicitly say that only in case the selection variable is defined in a OneDimBinning we don't allow it to be part of the MultiDimBinning too?

JoshHenryPeterson (approved):

PR #856, opened 2025-05-13
add charm correction stage and file with correction

marialiubarska:


thehrh:
As a heads up, the service seems to be missing a definition of the `expected_container_keys` (see https://github.com/icecube/pisa/blob/4.2/pisa/stages/xsec/nutau_xsec.py for instance). I hope that will also fix the failing service test.

marialiubarska:
I see, thank you, this makes it easier. I'll fix it later today

thehrh:
Sorry, I probably should've been more explicit: you still have to pass `expected_container_keys` to the init function of the `Stage` base class (see https://github.com/icecube/pisa/blob/4.2/pisa/stages/xsec/nutau_xsec.py#L65), just as we do with `expected_params`. It's also explained in the [service howto](https://github.com/icecube/pisa/blob/4.2/pisa/stages/how_to_service.md#the-__init__-method-definition).

Please also provide [a little more documentation](https://github.com/icecube/pisa/blob/master/pisa/stages/how_to_service.md#class-docstring) to make the service potentially useful for others: for example, is there a reference for this charm production bug in GENIE?

marialiubarska:
Sorry, that's my bad, forgot to actually pass it. I also added a more descriptive docstring explaining the purpose of the stage. There no outside references we can use for the bug (it was fixed in later versions of genie, but it was not reported in release notes), but I linked two of my presentations explaining the problem and showing use of the stage with VS

thehrh:
Excellent! The service test is also successful now.

pisa/stages/xsec/correct_charm_y.py
```diff
@@ -0,0 +1,163 @@
...
+class correct_charm_y(Stage):
+    """
+    This stage reweights inelasticity distribution
+    to remove the most obvious impact from charm
+    production bug in GENIE 2.12.8. The bug effectively 
+    kills most charm events.
+
+    Parameters:
+    -----------
+    correct_charm : bool
```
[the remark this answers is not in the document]
marialiubarska: It was mostly just for convenience, so the correction can be turned on/off in the existing distribution maker (e.g. when making comparison plots). The default is `False` because this stage is only meant for (approximate) bug impact testing and not anything else, so even if it somehow sneaks into actual analysis, the correction will be off by default

pisa/stages/xsec/correct_charm_y.py
```diff
@@ -0,0 +1,163 @@
...
+                ind_y = np.digitize(y, bins=yed)
+
+                # TODO: replace hardcoded number of bins?
+                ind_x[ind_x==31] = 30
+                ind_y[ind_y==31] = 30
+
+                res = np.zeros_like(lgE)
+                mask_upg = coszen<-0.9
+
+#                 if coszen<-0.9:
```
ts4051, line 96: Recommend removing commented out code when merging with main, if it is no longer needed

pisa/stages/xsec/correct_charm_y.py
```diff
@@ -0,0 +1,163 @@
...
+                    true_lg_energy = np.log10(container["true_energy"])
+                    true_y = container['bjorken_y']
+                    true_coszen = container['true_coszen']
+
+                    apply_mask = true_y >= 0
+#                     dis_mask = container['dis'] > 0
+#                     apply_mask = apply_mask * dis_mask
+                    container["apply_mask"] = apply_mask
+
+                    lgE_min = 0 #2.
```
ts4051, line 130: Another instance of commented out code to remove

pisa/stages/xsec/correct_charm_y.py
```diff
@@ -0,0 +1,163 @@
...
+
+                    container["charm_y_distr_corr"][valid_mask] = distr_valid_erange
+                    container["charm_y_distr_corr"][extrp_mask] = distr_extrap_erange 
+
+    def apply_function(self):
+        # modify weights
+        if self.correct_charm:
+            for container in self.data:
+                # reweighting all CC events
+                if container.name.endswith('_cc'):
```
ts4051, line 159: Should there also be a charm correction for NC events?

pisa/stages/xsec/correct_charm_y.py
```diff
@@ -0,0 +1,163 @@
...
+
+                    distr_valid_erange = eval_hist(true_lg_energy[valid_mask],
+                                                   true_y[valid_mask], true_coszen[valid_mask], nubar=is_nubar)
+                    distr_extrap_erange = eval_hist(np.ones_like(true_y[extrp_mask])*lgE_min,
+                                                    true_y[extrp_mask], true_coszen[extrp_mask], nubar=is_nubar)
+
+                    container["charm_y_distr_corr"][valid_mask] = distr_valid_erange
+                    container["charm_y_distr_corr"][extrp_mask] = distr_extrap_erange 
+
+    def apply_function(self):
```
ts4051, line 154: I'm a little surprised that this correction is being applied to all nu CC events, not just those with the GENIE charm flag set. Similarly, it seems to be equally applied to RES/QEL/DIS channels. Can you explain this please.

pisa/stages/xsec/correct_charm_y.py
```diff
@@ -0,0 +1,163 @@
...
+
+#                 if coszen<-0.9:
+#                     h_nucc = h_nucc_upg
+#                     h_nubarcc = h_nubarcc_upg
+#                 else:
+#                     h_nucc = h_nucc_oth
+#                     h_nubarcc = h_nubarcc_oth
+
+                if not nubar:
+                    res[mask_upg] = h_nucc_upg[ind_x[mask_upg]-1, ind_y[mask_upg]-1]
```
ts4051, line 104: Why is a different correction applied to upgoing events?

pisa/stages/xsec/correct_charm_y.py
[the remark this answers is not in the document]
thehrh: There is very little risk in my eyes of someone going through the list of services without looking at their documentation and just arbitrarily putting together a pipeline. As per my understanding, the PISA mechanism for not allowing a given service to affect your analysis pipeline has so far been to "leave it out". If we come to the conclusion that such a flag/switch is generally useful - for example to prevent having to re-initialise a distribution maker as in your scenario - it should be implemented generically at the `Stage` or `Pipeline` level in my opinion, such that it can apply to any service.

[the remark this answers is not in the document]
marialiubarska: Yeah, okay, I suppose it is not that dangerous to get rid of the flag. I'll remove it.

pisa/stages/xsec/correct_charm_y.py
marialiubarska: will remove comments

pisa/stages/xsec/correct_charm_y.py
marialiubarska: GENIE does not model interactions on heavy sea quarks, so all charm-production events are CC

pisa/stages/xsec/correct_charm_y.py
marialiubarska: The charm bug in GENIE is a missing coordinate system conversion, which results in unphysical momentum for charm events which are not directly upgoing (so are travelling along lab frame axis). These events are rejected during generation, which means we do not have any charm events except very upgoing region (which are still affected, but not a much). The largest effect we can predict from this is a change in the inelasticity distribution. This stage is designed to test only this first-order effect on an analysis. 

Since there are no charm events to reweight, we instead create pseudo-data where inelasticity distribution have been reweighted to match one from the correct simulation. This is only a test for particular effect from missing charm events, since without new simulation we can not correct the bug. Basically the best we can do is to verify that the bug does not affect our measurement.

pisa/stages/xsec/correct_charm_y.py
marialiubarska: This was a later modification to the test, which was introduced to account for the fact that some events in the very upgoing region manage to survive physical momentum check in genie. Because of this inelasticity distribution is affected differently in very upgoing events and a different correction ratio was produced.

marialiubarska:
hmm, okay, let me test it in an actual pipeline...

marialiubarska:
Ah right, it is failing because it expects container names to be from `nue_cc`, `nuebar_cc`, `numu_cc`, `numubar_cc`, `nutau_cc` or `nutaubar_cc`.
And [pisa_tests/test_services.py](https://github.com/icecube/pisa/blob/master/pisa_tests/test_services.py) tries to pass `test1_cc`. 
@thehrh do you have a suggestion on how to resolve it?

I also now tested the stage after all the changes and confirmed it still works

marialiubarska:
Sorry for the delay with this, I have modified the charm stage to use predefined lists of nu/nubar CC container keys, which can also be overwritten through stage arguments. Now the checks pass successfully

thehrh:
Thanks! If you want, you can now add an [`init_test` function](https://github.com/icecube/pisa/blob/4.2/pisa/stages/how_to_service.md#service-testing) at the bottom, which for example initialises the service like so:
```
def init_test(**param_kwargs):
    """Instantiation example that enables actual testing on toy containers from test_services.py"""
    return correct_charm_y(
        nu_cc_container_keys=["test1_cc"],
        nubar_cc_container_keys=["test2_nc"]
    )
```
to allow `pisa_tests/test_services.py` to execute and test the service's main code instead of just skipping both toy containers (`test1_cc` and `test2_nc`) which that testing script creates.

thehrh (commented):
I've left two more comments and will leave the final approval to @ts4051

JanWeldert:
I think if the two last comments from @thehrh are solved and @ts4051 gives his ok, this PR can be merged.

JanWeldert:
@ts4051 are you happy with this PR being merged?

JanWeldert (approved):

PR #857, opened 2025-05-23
Add Dockerfile

JanWeldert:
Add a docker file to enable the containerization of pisa. This can be used to freeze the software env at the end of an analysis and ensure reproducibility.


thehrh:
... in `[Documentation](...)` appears the same as the link to the actual PISA documentation at the top of the README. Could we just write `[here](...)` and add `(access restricted).` remark after that?

thehrh (approved):

PR #861, opened 2025-06-16
Merging minor changes developed as part of the oscNext DynEdge analysis development

ts4051:
1) Fix to collections module import
2) Fix copy module import in flux
3) Support relative flux gradients
4) Rename theta to theta23 in 2-flavor nu osc stage for easier switching between 3-nu and 2-nu


pisa/utils/hypersurface/hypersurface.py
```diff
@@ -27,9 +27,15 @@
...
 
 import os
-import collections
 import copy
 
+# Handle change over time in `collections` module
+from collections import OrderedDict
+try :
+    from collections.abc import Mapping, Sequence # Required as of py3.10
+except Exception as e :
```
thehrh, line 36: Why is there a need for this exception handling?

`Mapping` and `Sequence` are part of `collections.abc` in version [3.8](https://docs.python.org/3.8/library/collections.abc.html) - the lowest python version PISA currently supports. In fact, the `abc` module was introduced in version 3.3.

Furthermore, there are many other PISA modules which make use of the upper (`from collections.abc import Mapping, Sequence`) statement without exception handling.

Unless there's some particular motivation for it in this file, I'd suggest removing the try/except clause here.

pisa/stages/flux/mceq_barr.py
```diff
@@ -94,6 +95,7 @@ def __init__(
         table_file,
         include_nutau_flux=False,
         use_honda_nominal_flux=True,
+        use_relative_gradients=False,
```
thehrh, line 98: Documentation of this parameter is missing.

pisa/stages/flux/mceq_barr.py
```diff
@@ -287,47 +293,51 @@ def setup_function(self):
...
+            nu_flux_mceq[:, 1] = self.spline_tables_dict[arb_gradient_param_key]["numu" if nubar > 0 else "numubar"](
+                true_abs_coszen,
+                true_log_energy,
+                grid=False,
+            )
 
-                # numu(bar)
-                nu_flux_nominal[:, 1] = self.spline_tables_dict[arb_gradient_param_key]["numu" if nubar > 0 else "numubar"](
+            # nutau(bar)
+            # Currently setting to 0 #TODO include nutau flux (e.g. prompt) in splines
```
thehrh, line 326: This comment seems to relate to what the `include_nutau_flux` initialisation parameter effectuates. Could this be added to the service's docstring instead (which has unfortunately been missing both this and the `use_honda_nominal_flux` parameter for a while)?

thehrh:
Which bug does the MCEq Barr stage contain?

pisa/stages/osc/two_nu_osc.py
```diff
@@ -44,7 +44,7 @@ def __init__(self,
                 ):
 
         expected_params = (
-            'theta',
+            'theta23',
```
thehrh, line 47: The initialisation within the `init_test` function at the bottom of the module needs to reflect this naming change in order to get the service test to pass again.

ts4051:
@thehrh The flux stage bug was a missing import of `copy` iirc

pisa/stages/flux/mceq_barr.py
ts4051: Added now

pisa/utils/hypersurface/hypersurface.py
ts4051: Sounds good, have removed the older version

pisa/stages/osc/two_nu_osc.py
ts4051: Fixed

JanWeldert:
This looks good to me. The only suggestion I would have is to move the TODOs written as comments in the code into github issues. That way we can track them better.

thehrh (approved):
looks good

ts4051:
Thanks @thehrh and @JanWeldert

PR #863, opened 2025-06-17
Simple DIS correction

JanWeldert:
Add a stage to reweight DIS events based on a simple parameterization using bjorken variables. The parameterizations are hard coded for now. The stage simply evaluates the reweighting function and linearly interpolates between the correction and the original event weight.

### DIFF
--- a/pisa/core/binning.py
+++ b/pisa/core/binning.py
@@ -3070,10 +3070,9 @@ class VarBinning():
 
     If a `OneDimBinning` is passed to `selections`, the corresponding
     dimension must not be a dimension of any `MultiDimBinning` in `binnings`.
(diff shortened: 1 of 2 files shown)


JanWeldert:
I think with #865 this PR is not necessary anymore. Any thoughts?

CptBurtReynolds:
Given the outcome of Tom's study into decoupling the CSMS correction I agree with you. Teppei and I still have some non-grad students looking into some into genie-related stuff, but that's primarily with regards the differences pythia8 produces, and unrelated to this correction.

JanWeldert:
Alright, sounds good. I will close this PR then.

PR #864, opened 2025-06-20
Ultrasurfaces service: add option of only looking up gradients from neighbours among original (classification) event groupings

### LINKED ISSUE #782: Useless/misleading stage instance debug_mode attribute?
This attribute claims to disable caching at the [Stage](https://github.com/icecube/pisa/blob/master/pisa/core/stage.py) base class level, but it doesn't actually seem to have any functionality. No service throughout the code base implements any custom behaviour.
`debug_mode` was introduced in https://github.com/icecube/pisa/commit/4cb9172bd5a7b38c666f87f2bdeb92804a12d240 and used to have various effects as one can tell.

What to do with it? Make it, e.g., force `compute` always to call `compute_function` at least (no caching on identical parameter value hashes)?

### LINKED ISSUE #862: Service `discr_sys.ultrasurfaces`: ensure event category of nearest neighbor matches that of query
Description of the issue from https://github.com/icecube/pisa/blob/6015551ec42f879c251146de445147dce9a5e6fb/pisa/stages/discr_sys/ultrasurfaces.py#L159
```
At least in theory, this should always pick out the one exact event from the correct event category. If an event is not exactly matched, however, it's possible that the gradient for a numu_cc event might get picked from a nu_nc event, for example. We don't have any safeguards against that at this time, even though the information is in the DataFrame to do it.
```

Thich supposedly becomes particularly relevant when an event is encountered that wasn't part of the original training/classification problem.

thehrh:
Currently, with a set of USes for FLERCNN, the neutrino pipeline setup time increases from 64 s -> 69 s (`discr_sys.ultrasurfaces`: 30 s -> 35 s). Verification of pipeline outputs is coming soon (update: found complete agreement between distances and gradients from current master and those in this branch with `event_grouping_key = event_category` for two different nominal MC sets and their associated ultrasurface fit results).

Closes #862, closes #782

### DIFF
--- a/pisa/core/stage.py
+++ b/pisa/core/stage.py
@@ -44,12 +44,10 @@ class Stage():
 
     debug_mode : None, bool, or string
         If None, False, or empty string, the stage runs normally.
-
-        Otherwise, the stage runs in debug mode. This disables caching
-        (TODO: where or how?).
-        Services that subclass from the `Stage` class can then implement
-        further custom behavior when this mode is set by reading the value of
-        the `self.debug_mode` attribute.
+        Otherwise, the stage runs in debug mode: services that subclass
+        from the `Stage` class can then implement further custom behavior
+        when this mode is set by reading the value of the `self.debug_mode`
+        attribute.
 
     error_method : None or string (not enforced)
         An option to define one or more dedicated error calculation methods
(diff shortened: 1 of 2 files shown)


thehrh:
If this is merged, the `event_grouping_key`'s default value of "event_category" will apply to all existing pipeline configs containing the `discr_sys.ultrasurfaces` service. The original (i.e., potentially buggy) behaviour would have to be explicitly requested by setting the key to `None`. Whether the distance should be calculated in a different manner or how/if a tolerance should be set is still up for debate.

thehrh:
Closes #782

JanWeldert (approved):

thehrh:
KDtree query interface in v1.0: https://scikit-learn.org/1.0/modules/generated/sklearn.neighbors.KDTree.html?highlight=kdtree#sklearn.neighbors.KDTree.query

PR #872, opened 2025-08-26
Fix `pipeline._get_outputs` function for given output binning

### LINKED ISSUE #871: `Distribution_maker.get_outputs()` function gives wrong results with custom `output_binning`
The function doesn't rerun the distribution maker pipeline(s) with the custom `output_binning`, but tries to convert from the output binning defined in the config files to the given binning. Since this conversion is done in "average" mode it means that e.g. splitting a bin will lead to two bins with the same event count as the original bin. In other words, twice as many bins will give you twice as many events, which is obviously wrong.

JanWeldert:
If a binning is provided to the `_get_outputs` function, overwrite the hist stage apply mode and run the pipeline setup before actually running the pipeline. That way the pipeline will run with the new binning instead of trying to convert from its original (config file) binning to the given binning. 
This PR closes #871.


thehrh (changes requested):
see inline comments

thehrh (approved):

PR #874, opened 2025-09-05
Improved parameter handling of Map.plot() and fixed various issues

### LINKED ISSUE #869: Several parameters of map.plot() don't work
E.g. `logz` can't be used because we always set a `vmin` and `vmax`, `ax` only works if you also set `full_ax`, `title` doesn't do anything. Probably there are more. This function needs a cleanup.

bleaktwig:
Partially addresses #869.

Map.plot() crashes when `ax` is defined but `full_ax` is undefined due to bad logic ([here](https://github.com/icecube/pisa/blob/b4db437c1cda1db8f24283fdaa15c6f996a5c967/pisa/core/map.py#L724-L728)). This makes scripts such as [this example](https://github.com/icecube/pisa/blob/master/pisa_examples/IceCube_3y_oscillations_example.ipynb). This PR fixes the logic.


JanWeldert:
If `full_ax` should never be provided by a user, why is it an argument of the function then?

JanWeldert:
I think your second option is not unusual (also within pisa), but we can leave it as it is for now. Thanks for cleaning up that function.

JanWeldert (approved):

Issue #264, opened 2016-12-14
Map.py inconsistent handling of metrics / 


jllanfranchi:
inconsistent treatment of metrics: args handled differently across *chi2*, *llh*, and metric* methods of Map.py

LeanderFischer:
Ok, yeah, if this is still the case, then that should be tackled!

PR #881, opened 2025-10-15
Inconsistent treatment of metrics in pisa.core.map.Map and MapSet

bleaktwig:
This PR partially addresses #264.

* Made args handling consistent for all metrics in pisa.core.map.Map except for `generalized_poisson_llh()` (which fits a map against a model rather than against another map).
* Fixed a small bug in pisa.core.map.MapSet, where `MapSet.llh_total()` would call `.llh()` rather than `.llh_per_map()`, causing a TypeError.
* Fixed a bug where some metric methods would return nan if the input map had masked bins.
* Included "signed_sqrt_mod_chi2" in `CHI2_METRICS` in stats.py, which seems to have been left out due to an oversight.

This PR shouldn't change the behavior of already running code, except maybe for `barlow_llh()`, which used to handle Iterable objects separately from Map objects.


pisa/core/map.py
```diff
@@ -1568,241 +1568,92 @@ def split(self, dim, bin=None, use_basenames=False, pure_bin_names=False):
...
 
-        binned : bool
-
-        Returns
-        -------
-        total_llh : float or binned_llh if binned=True
+        # All metrics in utils.stats already check and return errors for nan
+        #     values, so any nan values that remain here are from incorrect
+        #     handling of bin masks bin masking.
+        if binned: return result
```
JanWeldert: Not a fan of indentations?

JanWeldert (commented):
This is a very nice PR that will remove a lot of unnecessary code. I just have a few comments.

JanWeldert (approved):

PR #883, opened 2025-10-22
rerun daemonflux example notebook

marialiubarska:
updated the example notebook after bugfix


JanWeldert (approved):
You could delete some of the larger outputs to reduce the file size a bit (the diff doesn't show up in the online version), but apart from that it looks good.

thehrh:
@marialiubarska @JanWeldert Did both of you verify the notebook outputs (https://github.com/icecube/pisa/blob/db24838f292d1f31a7d5a192c239e2e1f6318b44/pisa_examples/test_daemonflux_stage.ipynb) here?

3 years of DRAGON should not have summed (neutrino) count expectations limited to ~160, but rather ~700. And indeed, when I run the notebook on the current master branch, the maximum bin count is around that level. (Or compare the [IceCube_3y_oscillations_example.ipynb ](https://github.com/icecube/pisa/blob/db24838f292d1f31a7d5a192c239e2e1f6318b44/pisa_examples/IceCube_3y_oscillations_example.ipynb) notebook.)

Even my daemonflux/Honda ratio map looks different:
image

In general, no settings should have changed when it comes to the public data sample, which is used by this notebook, and neither should the `daemon_flux`/`honda_ip` services have undergone any changes affecting their default physics outputs, which makes me wonder how the results in this PR were produced.

I have a modified version of this notebook in the works in #930 (so no need for a dedicated PR fixing this).

PR #884, opened 2025-10-23
Update artificial PID changes

JanWeldert:
This PR:
1. Allows relative (artificial) pid improvements in the reco.resolutions stage. So far, only absolute changes were possible.
2. Removes the pid stage. Changes to the pid performance should be done in the reco stage in the same way as for other reconstruction quantities. The effects of the existing pid stage (simply scaling and shifting pid values) can also be achieved by changing the pid binning.

It would also be nice to be able to change the resolutions on the fly in an existing distribution maker, but I will leave that for a future PR.


thehrh (changes requested):
see discussion thread

thehrh (approved):

PR #885, opened 2025-10-29
Adding the possibility to choose the calibration file that daemonflux…

jpyanez:
Current daemonflux stage does not have the flexibility to define a calibration file. This adds that capability. If none is specified, the default calibration in the installed version of daemonflux is used.

### DIFF
--- a/pisa/stages/flux/daemon_flux.py
+++ b/pisa/stages/flux/daemon_flux.py
@@ -29,6 +29,9 @@ class daemon_flux(Stage):  # pylint: disable=invalid-name
     Parameters
     ----------
 
+    calibration_file: str
+        Path to the calibration file to be used
+    
     params: ParamSet
         Must have parameters: .. ::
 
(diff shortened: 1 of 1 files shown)


pisa/stages/flux/daemon_flux.py
```diff
@@ -198,7 +206,7 @@ def init_test(**param_kwargs):
     """Initialisation example"""
     param_set = []
     random_state = get_random_state(random_state=666)
-    for i, pname in enumerate(Flux(location='IceCube', use_calibration=True).params.known_parameters):
+    for i, pname in enumerate(Flux(location='IceCube', use_calibration=True, cal_file = self.cal_file).params.known_parameters):
```
JanWeldert: The `init_test` is not a function of the daemon_flux stage class, so it will not know the `self` reference. You can just pass `cal_file = None` for the test.

JanWeldert (changes requested):
This looks straight forward. Just the init test needs to be fixed.

JanWeldert (approved):

PR #889, opened 2025-12-12
update pylintrc file: generate minimal version from existing configuration

thehrh:
You might have noticed the many warnings emitted when running [pylint](https://github.com/pylint-dev/pylint). These are due to PISA's current pylintrc file, the last changes to which were made nearly seven years ago. At the time, Pylint 2.2.2 was the most recent release, whereas Pylint 4.0.4 was released a couple of weeks ago. The file was originally generated in https://github.com/icecube/pisa/commit/83e2832a6f267e03f41a13d9679b726dfb286ee1, possibly with Pylint 1.7, and only underwent three minor changes.

As a result, we are 
* specifying options that aren't recognised any longer:
```
E0015: Unrecognized option found: optimize-ast, files-output, no-space-check, bad-functions, function-name-hint, variable-name-hint, const-name-hint, attr-name-hint, argument-name-hint, class-attribute-name-hint, inlinevar-name-hint, class-name-hint, module-name-hint, method-name-hint (unrecognized-option)
```
* "useless" option values (no list provided here, but all warnings but one refer to https://github.com/pylint-dev/pylint/pull/4942, and the last, about removal of C0330, to https://github.com/pylint-dev/pylint/pull/3571) ,
* and unknown option values:
```
W0012: Unknown option value for '--enable', expected a valid pylint message and got 'old-octal-literal' (unknown-option-value)
W0012: Unknown option value for '--enable', expected a valid pylint message and got 'old-ne-operator' (unknown-option-value)
W0012: Unknown option value for '--enable', expected a valid pylint message and got 'long-suffix' (unknown-option-value)
```

The new pylintrc in this PR accounts for all the changes to Pylint's configuration options, and has been condensed down to a minimal version, containing only PISA's non-default option values. I produced it by generating a new file from the existing configuration with Pylint 4.0.4 (`pylint --generate-rcfile > <filename>`), a new file with defaults only (`pylint --rcfile="" --generate-rcfile > <filename2>`), and comparing the two.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -1,382 +1,235 @@
-[MASTER]
+# a minimal pylintrc file which only contains non-default settings (generated with pylint 4.0.4)
+[MAIN]
 
-# Specify a configuration file.
-#rcfile=
+# List of module names for which member attributes should not be checked and
+# will not be imported (useful for modules/projects where namespaces are
+# manipulated during runtime and thus existing member attributes cannot be
+# deduced by static analysis). It supports qualified module names, as well as
+# Unix pattern matching.
+ignored-modules=numpy,scipy,scipy.stats,matplotlib,matplotlib.cm,uncertainties.unumpy
 
-# Python code to execute, usually for sys.path manipulation such as
-# pygtk.require().
-#init-hook=
-
-# Add files or directories to the blacklist. They should be base names, not
-# paths.
-ignore=CVS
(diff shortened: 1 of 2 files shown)


thehrh:
Finally, concerning the three unknown option values (W0012) above, they all seem to have been relevant during a Python 2 -> 3 transition, so there's nothing to do with respect to those either.

JanWeldert (approved):

PR #895, opened 2026-01-06
Add Dependabot configuration (pip + github-actions)

thehrh:
I propose testing if and how Dependabot can help us with our dependency updates. The suggested file has been produced using copilot. Once it is merged, we should be able to trigger a manual first Dependabot run from the dependency graph.
  
* Adds .github/dependabot.yml to enable Dependabot version updates for pip (setup.py/setup.cfg) and GitHub Actions workflows.
* Schedule: weekly
* Notes: Git-based dependencies (VCS URLs) and conda environment files are not managed by Dependabot (need pinned requirements or pyproject.toml for improved updates)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"                    # setup.py and setup.cfg are in repository root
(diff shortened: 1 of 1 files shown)


bleaktwig (approved):
Good idea!

PR #914, opened 2026-03-06
Share pipeline output binning through new `ContainerSet` global auxiliary data (+ related changes)

JanWeldert:
- Add global auxiliary data dict to container set, to make it possible to share keys between different containers. This can be used to share the output binning and regularized output binning.
- Check now if the output binning is the same as the apply mode of the hist stage(s).
- hist and snowstorm_hist stage can have apply_mode = None. In that case they will directly take the output binning.

Using the regularized output binning from the hist stage in the snowstorm hist stage speeds up the latter's computation.

It is important to enforce the apply modes of the different stages (especially the hist stage) to be the same as the output binning because re-binning is not currently supported by PISA. In principle, having a different hist stage apply mode and output binning would be possible and produce the correct results. However, it would not make sense to bin the events in the hist stage just to re-bin them immediately in order to obtain the output map.


pisa/stages/cont_sys/snowstorm_hist.py
```diff
@@ -149,9 +155,19 @@ def compute_function(self):
             # Only need per event infos if we want to calculate the gradients
             if calc_grads:
                 container.representation = self.calc_mode
-                sample = np.array([container[name] for name in self.apply_mode.names])
                 syst = [container[sys] for sys in self.systematics]
                 weights = container["weights"]
+                sample = []
+                dims_log = [d.is_log for d in self.apply_mode]
+                dims_ire = [d.is_irregular for d in self.apply_mode]
+                for dim, is_log, is_ire in zip(self.data["regularized_output_binning"], dims_log, dims_ire):
```
thehrh, line 168: "regularized_output_binning" is similar to an entry of "expected_container_keys", but now corresponds to an auxiliary global ContainerSet key. While there is no mechanism for declaring whether a service requires one of these, it may be fine to keep it that way for now (as there is no use case yet other than histogramming followed by this service).

pisa/stages/cont_sys/snowstorm_hist.py
```diff
@@ -149,9 +155,19 @@ def compute_function(self):
             # Only need per event infos if we want to calculate the gradients
             if calc_grads:
                 container.representation = self.calc_mode
-                sample = np.array([container[name] for name in self.apply_mode.names])
                 syst = [container[sys] for sys in self.systematics]
                 weights = container["weights"]
+                sample = []
```
thehrh, line 165: Mention that you copied the following lines from the utils.hist service.

pisa/stages/utils/hist.py
```diff
@@ -116,12 +119,10 @@ def setup_function(self):
                     dimensions.append(new_dim)
                 else:
                     dimensions.append(dim)
-            self.regularized_apply_mode = MultiDimBinning(dimensions)
+            self.data["regularized_output_binning"] = MultiDimBinning(dimensions)
             logging.debug(
-                "Using regularized binning:\n" + str(self.regularized_apply_mode)
+                "Using regularized binning:\n" + str(self.data["regularized_output_binning"])
```
thehrh: use [lazy logging](https://medium.com/flowe-ita/logging-should-be-lazy-bc6ac9816906)

pisa_examples/resources/settings/pipeline/example.cfg
```diff
@@ -45,7 +45,6 @@ output_key = weights, errors
 #   2) apply_mode   : the data representation that is by default selected before
 #                     the stages `apply_function` is executed
```
thehrh, line 46: use comment 1) above to explain why this particular stage config does not specify `calc_mode`

pisa_examples/resources/settings/pipeline/fast_example.cfg
```diff
@@ -45,7 +45,6 @@ output_key = weights, errors
 #   2) apply_mode   : the data representation that is by default selected before
 #                     the stages `apply_function` is executed
```
thehrh, line 46: dito (just copy & paste)

thehrh (changes requested):
Just a handful of small remarks to address.

It will be important to adapt and improve (where necessary) the two notebook tutorials on [stage modes](https://github.com/icecube/pisa/blob/12ce31565757f5b4835976ba1acd54311c297969/pisa_examples/pisa_modes.ipynb) and [eventwise information](https://github.com/icecube/pisa/blob/12ce31565757f5b4835976ba1acd54311c297969/pisa_examples/event_info.ipynb) in an upcoming PR.

thehrh (approved):
I just fixed the one outstanding review comment. Looks good to me now.

Issue #661, opened 2021-03-15
surprising behavior from Analysis.fit_hypo 


atfienberg:
After setting a distribution maker's parameters to new `Param` objects, `fit_hypo` may discard these parameters before fitting. One example of how this could occur is this:

1. Conduct a fit
2. Set the distribution maker's parameters to `copy.deepcopy(best_fit_params)`
3. Evaluate the distribution at the best fit parameters
4. Fix some parameters to new values 
5. Run another fit

Under some conditions, the fit in step 5. will discard the parameters from step 4. This occurs because of the call to `hypo_maker.select_params(full_param_selections)`. I found this behavior very surprising. Whether the fit actually uses parameters fixed before the call to `fit_hypo` depends on whether the user modifies parameters in place (which might work) or sets the parameters to new `Param` objects (which probably won't work),

philippeller:
Related to #644

mvprado0:
I am finding the same problem, and with the current Pisa functions for ParamSet, there doesn't seem to be a way to make the current analysis.py code use best fit parameter starting values and set them free at the same time. All one can do if you want to keep the best fit values is fix the parameters in place as described above. A quick and simple way to fix this without messing with `select_params()` or `ParamSet` would be with a simple flag in `fit_hypo` that incorporates an if statement into the line `hypo_maker.select_params(full_param_selections)`. 

Something like: 

`def fit_hypo(self, data_dist, hypo_maker, metric, minimizer_settings, using_previously_chosen_params=False, ...):`
`If using_previously_chosen_params==False:`
`hypo_maker.select_params(full_param_selections)`

If this is okay, I would be happy to make a PR for it.

LeanderFischer:
[...]

Issue #644, opened 2020-10-14
Analysis class


philippeller:
Current analysis class is overly complicated with 2k+ SLOCs and nested function calls that I cannot really see through....
Simpler >> Complicated

thehrh:
Keeping this open because after #916 the analysis module still has around ~3.5k LOC. However, ~700 of these constitute tests and ~400 make up the `Analysis` class docstring, which simultaneously currently serves as the fitting tutorial.

PR #916, opened 2026-03-09
Refactor analysis package

### LINKED ISSUE #827: Reduce memory and disk-space usage of `HypoFitResult`
* an analyser making use of https://github.com/icecube/pisa/blob/dacc631a04821003aa012e74908bda689c5936c3/pisa/analysis/analysis.py#L963 is unable to switch off recording and storing of fit/minimiser history, best-fit asimov template, detailed metric maps
* depending on nature of analysis, this can give rise to huge memory footprint, and, if no manual purging is done before storing, may take up unnecessary disk space

* unfortunately, there is little documentation explaining the design choices behind the [`HypoFitResult` class](https://github.com/icecube/pisa/blob/dacc631a04821003aa012e74908bda689c5936c3/pisa/analysis/analysis.py#L443) and its integration into the fitting functions (e.g., it's not obvious to me that the deserialisation mechanism should require all attributes in `_state_attrs` to be present)

* setting `blindness > 0` suppresses storing `fit_history`, but would have other unintended side effects for an analyser only interested in reducing analysis resource usage

* introduce a parameter to the call to `fit_recursively` which will force the fit to omit the most detailed information?

### LINKED ISSUE #910: `HypoFitResult` does not ensure consistency between metric value and the pair of data distribution & hypo maker
It is currently possible to create [such an instance](https://github.com/icecube/pisa/blob/57d7cfacea5bf9c1569778255c953376ac1fafdb/pisa/analysis/analysis.py#L616) in an inconsistent state: the metric value calculated from `data_dist` and `hypo_maker` does not need to agree with the externally computed `metric_val`.

Such a state is unlikely to be desired and should arguably result in an exception, but at least a warning.

### LINKED ISSUE #913: Non-deterministic success/failure of constrained slsqp test minimizer
The assertion https://github.com/icecube/pisa/blob/12ce31565757f5b4835976ba1acd54311c297969/pisa/analysis/analysis.py#L3981 has begun to randomly cause the associated test of constrained scipy slsqp minimization to fail, as observed for example in [this workflow run](https://github.com/icecube/pisa/actions/runs/22438615054/job/64974899072).

One would naively expect the outcome to be deterministic though, given the deterministic seed and the lack of any other non-deterministic settings.

We could consider catching the exception, simultaneously stating that it is under investigation and referring here, or remove the test.

### LINKED ISSUE #915: scipy.optimize DeprecationWarning: The `disp` and `iprint` options of the L-BFGS-B solver are deprecated and will be removed in SciPy 1.18.0.
To be ready for the next minor SciPy release, these options therefore need to be removed from [these](https://github.com/icecube/pisa/blob/12ce31565757f5b4835976ba1acd54311c297969/pisa/analysis/analysis.py#L3734) test settings and our [default-setting function](https://github.com/icecube/pisa/blob/12ce31565757f5b4835976ba1acd54311c297969/pisa/analysis/analysis.py#L221).

See e.g. https://docs.scipy.org/doc/scipy-1.17.0/reference/optimize.minimize-lbfgsb.html for the removal announcement.

thehrh:
* move configuration of scipy and nlopt minimization settings to separate modules
* move parameter manipulation tools to another
* make fit_history and detailed metric info in `HypoFitResult` configurable (less memory and disk space usage by default) and verify consistency between detailed metric info and the external metric value (where possible; not for a fit that used `Detectors` -> TODO added)
* catch any exception due to random failure of constrained-slsqp test
* remove setting of iprint and disp options for l-bfgs-b
* cosmetics
* removal of unused & broken `Analysis` class and rename `BasicAnalysis -> Anaysis`, but keeping an alias (is a replacement of nofit_hypo desirable?)
* remove global scipy minimization from automatic unit testing (for reduced workflow duration)
* finally document scipy fitting possibilites in `Analysis` class docstring
* use [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) extension to link from PISA docs to those of other projects within Python ecosystem -> install [intersphinx-registry](https://pypi.org/project/intersphinx-registry/) package to keep URLs updated when "develop" dependency group is selected
* & a few other improvements

Fixes #915
Fixes #913 (create a replacement issue for the randomness) 
Closes #910 
Fixes #827
Should at least partially address #644

In addition, #661 could be closed as not planned if this PR goes through.


pisa/analysis/analysis.py
```diff
@@ -1,15 +1,13 @@
 """
-Common tools for performing an analysis collected into a single class
-`Analysis` that can be subclassed by specific analyses.
+Common tools for performing an analysis collected into the class
+`BasicAnalysis` that can be subclassed by specific analyses.
```
JanWeldert: Wouldn't it be good to rename the `BasicAnalysis` class to `Analysis` now that the `Analysis` class is gone? That way other scripts could still `import Analysis`.

pisa/analysis/analysis.py
```diff
@@ -733,12 +278,53 @@ def detailed_metric_info(self):
...
+            metric_val_from_maps = np.sum(
+                [d[self.metric[0 if not is_detectors else i]]["maps"]["total"] for
+                 i, d in enumerate(self._detailed_metric_info)]
+            )
+            if is_detectors:
+                # 1)
+                # TODO: We don't have access to the Detectors instance itself here
+                # -> no straightforward way to correctly determine the total prior
+                # contribution (cf. Detectors.init_params())
+                metric_val_from_priors = np.nan
```
JanWeldert, line 301: Wouldn't `total_metric_val_from_detailed` always be `nan` if `is_detectors`?

JanWeldert (changes requested):

pisa/analysis/analysis.py
thehrh: yes, and in that case no consistency check will and can be done

pisa/analysis/analysis.py
thehrh: I'd tend to agree with the sentiment that "Analysis" is a more appropriate name (the single fit it returns might be considered "basic" compared to a set of fits as in a scan, but the available fitting methods are anything but "basic"). We can define the alias `BasicAnalysis = Analysis` for backwards compatibility.

JanWeldert (approved):

PR #919, opened 2026-03-16
Add fast workflow for building PISA docker image and publishing to GitHub packages

thehrh:
Note that the dockerfile uses [uv](https://github.com/astral-sh/uv) instead of conda, which cuts down the image build duration to approx. 1/3 (3 mins.) - even faster than if we replaced conda by mamba. Also, the resulting image is roughly only half as large (1.75 GB) as the one produced by the current dockerfile.

I have tested the general setup with this workflow + dockerfile in my repo, downloaded the image via `docker pull`, and started a container.

Here, the image from a _manual_ workflow dispatch would have to be downloaded via the command  `docker pull ghcr.io/icecube/pisa:master` if I'm not mistaken. When triggered by a release, master should be replaced by the corresponding release tag.

### DIFF
--- a/.github/workflows/release_docker.yml
+++ b/.github/workflows/release_docker.yml
@@ -1,0 +1,60 @@
+# Workflow adapted from https://docs.github.com/en/actions/tutorials/publish-packages/publish-docker-images#publishing-images-to-github-packages
+name: Create and publish a Docker image
+
+# Run when a release is created or when triggered manually
+on:
+  release:
+    types: [created]
+  workflow_dispatch:
+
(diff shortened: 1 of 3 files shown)


JanWeldert:
Would it make sense to add one or two sentences to the [README](https://github.com/icecube/pisa/blob/master/README.md#containerization)?

thehrh:
Done. Anything else? We'll have to see whether the images indeed show up at the link specified in the README once we've dispatched the workflow.

JanWeldert (approved):

PR #920, opened 2026-04-01
Overhaul sphinx docs

thehrh:
The [current PISA html docs](http://icecube.github.io/pisa) only contain the auto-generated Python API references/indexes.

In addition to introducing a different theme, this PR includes various of our own resources (selection of important markdown files, jupyter notebooks) in the html docs, as well as a changelog generated from PISA's [releases](https://github.com/icecube/pisa/releases). New sphinx extensions need to be installed to achieve this, as the modified list of develop packages in setup.py reveals.

I've uploaded the docs built from this branch to https://user-web.icecube.wisc.edu/~tehrhardt/pisa_docs so that you can take a look first.


thehrh:
In my opinion, it isn't too important to have the automatically generated reST files (docs/source/pisa.stages.xsec.rst and so on, by sphinx-apidoc, which for some reason now produces a changed directive order) in here&mdash;or to keep them updated as code structure changes:
* the [documentation workflow](https://github.com/icecube/pisa/blob/beed10498f22b22cae96e63658473dea810be3d5/.github/workflows/documentation.yml) generates an up-to-date list of reST files before building the HTML docs anyway
* anyone building the docs locally on their machine can quickly run sphix-apidoc themselves also; only if they don't and code structure has changed compared to whatever the list of reST files in the master branch represents will they end up with outdated sphinx docs

JanWeldert (approved):

PR #936, opened 2026-05-15
Tolerances for snowstorm additional params

JanWeldert:
Allows a user to set tolerances for the parameters of the snowstorm stage that require a re-calculation of the systematics gradients (e.g. osc params). Setting no tolerance results in a re-calculation every time the parameter is changed.

### DIFF
--- a/pisa/stages/cont_sys/snowstorm_hist.py
+++ b/pisa/stages/cont_sys/snowstorm_hist.py
@@ -6,12 +6,11 @@ The method is based on this paper: https://arxiv.org/pdf/1909.01530
 
 import numpy as np
 
-from pisa import ureg
+from pisa import ureg, FTYPE
 from pisa.core.binning import MultiDimBinning
 from pisa.core.param import Param, ParamSet
 from pisa.core.stage import Stage
 from pisa.core.translation import histogram
-from pisa.utils.comparisons import FTYPE_PREC
 
 __all__ = [
     "snowstorm_hist",
@@ -62,9 +61,15 @@ class snowstorm_hist(Stage):  # pylint: disable=invalid-name
         Parameters that are no detector systematics but if changed require a
         re-calculation of the gradients (e.g. osc params).
     tolerances : list of float (default None)
-        Tolerances for 'additional_params'. If a parameter is changed by more than
-        its tolerance during a fit, the gradients are re-calculated. 'None' will
-        result in a tolerance that corresponds to the machine precision.
+        Numerical tolerances for `additional_params`, to be given in the same order.
+        If none are given (default), all gradients will be updated each time at least
+        one parameter value changes (most accurate but slowest option). If tolerances
+        *are* given, though, and *all* parameters move away from their initial values
+        by *less* than the specified tolerances over the course of a single fit, gradients
+        will *never* be re-calculated. Only once at least one parameter's overall delta
+        exceeds its corresponding tolerance the gradients will be updated (and a new
+        reference point set). Subsequent modfications of the parameter vector will
+        then be compared to this updated referenced point.
     params : ParamSet
         Note that the params required to be in `params` are those listed in 
         `systematics` plus those listed in `additional_params`.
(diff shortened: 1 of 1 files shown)


pisa/stages/cont_sys/snowstorm_hist.py
```diff
@@ -105,6 +111,16 @@ def __init__(
...
 
+        if isinstance(tolerances, str):
+            self.tol = eval(tolerances)
+        elif tolerances is None:
+            self.tol = [FTYPE_PREC] * len(self.additional_params)
+        else:
+            self.tol = tolerances
+        assert isinstance(self.tol, list)
+        assert len(self.tol) == len(self.additional_params)
+        self.tol = np.array(self.tol)
```
thehrh: * instantiate array with appropriate data type: `self.tol = np.array(self.tol, dtype=FTYPE)`
* `from pisa import ureg, FTYPE` at top

pisa/stages/cont_sys/snowstorm_hist.py
```diff
@@ -146,8 +162,8 @@ def setup_function(self):
     def compute_function(self):
         # First check if we need to calculate the gradients or if we can use the already stored ones.
         # We need to calculate if an additional params value or the apply_mode changed
-        additional_params_values = [self.params[p].m for p in self.additional_params]
-        if additional_params_values != self.additional_params_values:
+        additional_params_values = np.array([self.params[p].m for p in self.additional_params])
```
thehrh: instantiate with `dtype=FTYPE`

pisa/stages/cont_sys/snowstorm_hist.py
```diff
@@ -57,9 +58,13 @@ class snowstorm_hist(Stage):  # pylint: disable=invalid-name
     simulation_dists_params : list of tuples of floats
         Parameters of the simulation distributions. (mean, std) for 'gauss' and
         (min, max) for 'uniform'.
-    additional_params : list of str
+    additional_params : list of str (default None)
         Parameters that are no detector systematics but if changed require a
         re-calculation of the gradients (e.g. osc params).
+    tolerances : list of float (default None)
```
thehrh, line 63: This description is a bit brief for my liking. I would reformulate this for clarity, for instance like so (adapt if necessary):
```
Numerical tolerances for `additional_params`, to be given in the same order.
If none are given (default), all gradients will be updated each time at least
one parameter value changes (most accurate but slowest option). If tolerances
*are* given, though, and *all* parameters move away from their initial values
by *less* than the specified tolerances over the course of a single fit, gradients
will *never* be re-calculated. Only once at least one parameter's overall delta
exceeds its corresponding tolerance will all gradients be updated (and a new
reference point set). Subsequent modfications of the parameter vector will
then be compared to this updated referenced point.
```

pisa/stages/cont_sys/snowstorm_hist.py
```diff
@@ -105,6 +111,16 @@ def __init__(
             self.additional_params = additional_params
         assert isinstance(self.additional_params, list)
 
+        if isinstance(tolerances, str):
+            self.tol = eval(tolerances)
+        elif tolerances is None:
+            self.tol = [FTYPE_PREC] * len(self.additional_params)
```
thehrh: Having considered this a bit more, instantiating `self.tol = np.zeros_like(self.additional_params, dtype=FTYPE)` is probably more prudent. We want to err on the side of caution here. You are also not checking parameters' rescaled values in `compute_function`, so all values could have vastly different magnitudes. I'd say just update gradients and the reference point once *any* change is detected, however small it may be and irrespective of `FTYPE_PREC`.

thehrh (changes requested):
See the handful of inline comments.

NB: From a design standpoint, attributes such as `grads`, `tol`, `additional_params_values` etc. should better be declared as non-public (`self._tol` etc.) IMO, to indicate that they are only meant for internal use, in particular because this service has a custom caching mechanism. I'm aware that other services don't adhere to such a design, so approval won't depend on this.

thehrh (approved):

PR #944, opened 2026-05-29
ftype and format for range

JanWeldert:
Specify ftype for parameter ranges and make sure they are shown as floats when tabulated.

### DIFF
--- a/pisa/core/param.py
+++ b/pisa/core/param.py
@@ -894,6 +894,8 @@ class ParamSet(MutableSequence, Set):
                 table.append([p.name, p.value, p.nominal_value, p.range, p.prior, p.units, p.is_fixed])
             else:
                 if p.range is not None:
+                    # convert range to float values that will be displayed nicer than e.g. numpy floats.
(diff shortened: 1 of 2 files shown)


thehrh:
Please add comments to the code that explain *why* you chose to implement this/what this achieves. The motivation is currently missing both from the description of this pull request and, more importantly, from the code itself. Otherwise, future maintenance is needlessly complicated.

thehrh:
Note that this means that integer range bounds (or rather their dimensionless magnitudes) in config files will now be consistently converted into floating-point numbers with a precision corresponding to PISA's global `FTYPE`. I'm not aware of any disadvantages and the success of none of the unit tests seems to depend on integer range bounds either. 

For example, before, one might have specified one of the bounds as an integer and the other as a float, and the `range_` array would simply have been created with `dtype=np.float64`. The new behaviour removes dtype dependence on such flukes (though ranges can still be manually overridden of course).

thehrh (approved):

PR #947, opened 2026-06-05
Move detectors params handling completely in detectors class

### LINKED ISSUE #932: `analysis.fit_recursively()` fails when running over the Detectors class
When creating an instance of the `Detectors` class using two different pipelines, the method `analysis.fit_recursively(...)` fails, throwing: `IndexError: pop from empty list`. A minimal replicating example and full traceback are included in the discussion.

### LINKED ISSUE #946: Detectors class doesn't update parameters with the same name that are not shared in fit
Currently, parameters that have the same name in two distribution_makers but are not shared will get the name of the detectors added (starting from the second detector). These parameters with detector names in it don't get updated in a fit.

JanWeldert:
This PR

- resolves #946 
- streamlines the detectors class test
- moves the handling of param updates completely in the detectors class. Every time an output is generated, it is checked if a parameter changed. If so the parameters in the individual distribution makers are updated. Each detectors class function now also initializes the params so their changes to the distribution maker params are propagated.

### DIFF
--- a/pisa/core/detectors.py
+++ b/pisa/core/detectors.py
@@ -355,9 +355,7 @@ class Detectors(object):
                 self._distribution_makers[i]._set_rescaled_free_params(rp)
 
 
-def test_Detectors(verbosity=Levels.WARN):
-    from pisa.analysis.analysis import update_param_values_detector
-
+def test_Detectors():
     """Run a combination of two DeepCore detectors."""
     p1_nu = Pipeline("settings/pipeline/IceCube_3y_neutrinos.cfg")
     p1_mu = Pipeline("settings/pipeline/IceCube_3y_muons.cfg")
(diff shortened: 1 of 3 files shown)


pisa/core/detectors.py
```diff
@@ -60,28 +60,28 @@ def __init__(self, pipelines, label=None, set_livetime_from_data=True, profile=F
...
-            self._distribution_makers[i] = DistributionMaker(pipelines=pipelines, 
+            self._distribution_makers[i] = DistributionMaker(pipelines=pipelines,
                                                              set_livetime_from_data=set_livetime_from_data,
                                                              profile=profile
                                                             )
-            
+
         for sp in self.shared_params:
             N, n = 0, 0
             for distribution_maker in self._distribution_makers:
```
thehrh: You have made 13 substitutions `self -> self._distribution_makers` even though the class implements `__iter__()` (https://docs.python.org/3/reference/datamodel.html#emulating-container-types).
The only reason I could see for this change now would be performance impact due to iteration creation whenever you loop over the contained `DistributionMaker`s (time difference between `for d in self:` and `for d in self._distribution_makers`:). While the relative difference could be large because `self._distribution_makers` typically only has a couple of entries, the absolute difference would probably be so tiny that it wouldn't be relevant in practice.

pisa/core/detectors.py
```diff
@@ -172,71 +177,76 @@ def update_params(self, params):
...
     @property
     def distribution_makers(self):
         return self._distribution_makers
 
     @property
     def params(self):
         return self._params
-    
+
+    def get_params_hash(self):
```
thehrh: `ParamSet` implements its own `hash`. Why not the following?
```
@property
def params_hash(self):
    return self._params.hash
```

pisa/core/detectors.py
```diff
@@ -157,12 +157,17 @@ def get_outputs(self, **kwargs):
         List of MapSets if `return_sum=True` or list of lists of MapSets if `return_sum=False`
 
         """
+        new_params_hash = self.get_params_hash()
+        if new_params_hash != self.params_hash:
```
thehrh: Given a `params_hash` property, I believe this block would read:
```
new_params_hash = self.params_hash
if new_params_hash != self._params_hash:
    self.update_params(self.params, init_params=False)
    self._params_hash = new_params_hash
```

pisa/core/detectors.py
```diff
@@ -172,71 +177,76 @@ def update_params(self, params):
...
-                elif param.name in params.names: # two parameters with the same name but not shared 
+                elif param.name in params.names: # two parameters with the same name but not shared
                     # add detector name to the parameter name
                     changed_param = deepcopy(param)
                     changed_param.name = param.name + '_' + distribution_maker.detector_name
                     params.extend(changed_param)
                 else:
                     params.extend(param)
         self._params = params
+        self.params_hash = self.get_params_hash()
```
thehrh: Given a `params_hash` property, I believe this line would read
`self._params_hash = self.params_hash`.

pisa/core/detectors.py
```diff
@@ -93,7 +93,7 @@ def __init__(self, pipelines, label=None, set_livetime_from_data=True, profile=F
                 raise NameError(f'Shared param {sp} only exists in {N} detectors.')
             if n > 0 and n != N:
                 raise NameError(f'Shared param {sp} exists in {N} detectors but only a free param in {n} detectors.')
-        
+
```
thehrh, line 96: Given a `params_hash` property, set `self._params_hash=None` here and add an attribute docstring to make class design more transparent?
```
self._params_hash = None
"""Hash value of `ParamSet` constructed by last `init_params()` call"""
```

pisa/core/detectors.py
```diff
@@ -157,12 +157,17 @@ def get_outputs(self, **kwargs):
...
+            self.update_params(self.params, False)
+            self.params_hash = new_params_hash
+
         outputs = [distribution_maker.get_outputs(**kwargs) for distribution_maker in self]
         return outputs
 
-    def update_params(self, params):
-        if isinstance(params,Param): params = ParamSet(params)
-            
+    def update_params(self, params, init_params=True):
```
thehrh, line 170: Missing docstring, please provide one.

pisa/core/detectors.py
```diff
@@ -172,71 +177,76 @@ def update_params(self, params):
...
     @property
     def params(self):
         return self._params
-    
+
+    def get_params_hash(self):
+        return hash(self._params)
+
     def init_params(self):
         """Returns a ParamSet including all params of all detectors. First the shared params
```
thehrh: Doesn't "return"

pisa/core/detectors.py
```diff
@@ -354,10 +368,10 @@ def _set_rescaled_free_params(self, rvalues):
                     rp.insert(spi[i][j][0], sp[spi[i][j][1]])
                 self._distribution_makers[i]._set_rescaled_free_params(rp)
 
+        self.init_params()
 
-def test_Detectors(verbosity=Levels.WARN):
-    from pisa.analysis.analysis import update_param_values_detector
 
+def test_Detectors():
```
thehrh, line 384: Can you come up with a test which demonstrates what you fixed (not least because the notebook was unable to prevent #946)?

pisa_examples/Multi_detector_test.ipynb
```diff
@@ -458,7 +458,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "When checking the chi2 value from the fitted model, you maybe see that it is around 113, while in the minimizer loop we saw it converged to 116. It is important to keep in mind that in the fit we had extended the metric with prior penalty terms. When we add those back we get the identical number as reported in the fit."
+    "When checking the chi2 value from the fitted model, you maybe see that it is around 111, while in the minimizer loop we saw it converged to 114. It is important to keep in mind that in the fit we had extended the metric with prior penalty terms. When we add those back we get the identical number as reported in the fit."
```
thehrh, line 421: Relating to this notebook, it really doesn't need cells such as these, which were originally taken from the public 3-yr example notebook:
```
The two pipelines are quite different, with most complexity in the neutrino pipeline, that has several Stages and free parameters
```
or
```
While the muon pipleine is rather simple
```
etc.

Basically, all the content that has nothing to do with the `Detectors` functionality should be removed. Instead, just include a link to the public 3-yr example notebook at the top (for details about how pipelines work).

Also, you could add more instructions and hints with respect to working with the `Detectors` class in this notebook (setting parameters correctly, when are changes propagated, ...). A more tutorial-like notebook would be really useful to include in the sphinx docs (https://icecube.github.io/pisa/docs/tutorials.html).

thehrh (changes requested):
As a final remark, if `update_param_values_detector` should remain (still used a few times in analysis.py after all), it would benefit from a more detailed docstring than
```
"Modification of the update_param_values function to use with the Detectors class"
```
, in particular, why it is needed and when to use it (compare explanation in `update_param_values`).

pisa/core/detectors.py
thehrh: Ok, turns out such a property (or getter function) is unnecessary here and perhaps better when not exposed to the user (but the remark on the hash property of `ParamSet` still holds of course).

thehrh:
Are you done? I could live with leaving the `update_param_values_detector` request and possibly some more `Detectors` usage examples to a follow-up PR.

thehrh:
How would you proceed with #932? If I understand correctly, this will now render unnecessary the workaround you described there, because the state of the attached `ParamSet` would be compared  to that with which the `Detectors` instance was initialised at some point before the minimisation starts/before the first template is computed?

JanWeldert:
resolves #932

thehrh (approved):

PR #963, opened 2026-07-02
MEC sys stage

ts4051:
Added a simple MEC event systematic uncertainty stage, which scales the normalization.

### DIFF
--- a/pisa/stages/xsec/mec_sys.py
+++ b/pisa/stages/xsec/mec_sys.py
@@ -1,0 +1,73 @@
+"""
+Systematic uncertainty model for Meson Exchange Current (MEC) events.
+"""
+
+from pisa import ureg
+from pisa.core.param import Param, ParamSet
+from pisa.core.stage import Stage
+from pisa.utils.profiler import profile
+
+__all__ = ["mec_sys", "init_test"]
+
+
+class mec_sys(Stage):  # pylint: disable=invalid-name
+    """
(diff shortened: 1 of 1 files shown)


JanWeldert (approved):

thehrh:
@JanWeldert There seems to be a new major numba release, which might be causing `test_fbwkde` to get stuck? There's already a comment there indicating some numba dependence of the test outcome: [ `# tests failing with numba 0.62.1`](https://github.com/icecube/pisa/blob/902322bcf3e821a5f7a07b019e1fd7ab35329b25/pisa/utils/vbwkde.py#L626 ).

I'm going to try reproducing this with a fresh local install.

thehrh:
I had to manually terminate the test with numba [0.66.0](https://numba.readthedocs.io/en/stable/release/0.66.0-notes.html#version-0-66-0-30-june-2026). After downgrading to numba 0.65.1, it rapidly completed (though still with the 'division by zero' problem).

JanWeldert:
Do we still use this module somewhere? Otherwise we can skip the test and solve the problem once we use the service again.

thehrh:
> Do we still use this module somewhere? Otherwise we can skip the test and solve the problem once we use the service again.

That might be the way to go for now. The module is only being used by the unit testing script. And all three test functions are already effectively deactivated by catching the "division by zero" exception.

